PR #8, opened 2021-03-04
json: normalized using jq

hunchback:
how can I check that the resulting json's actually work ?


roytman:
We don't need this formating. These files (at least that we need) are coming from OVS/OVN sources.

roytman:
PR#9 is a replacement.

Issue #16, opened 2021-03-07
fixme - decoder not working on `[]string` type fields.


OmerBenHayun:
## the problem:
A field that need to decoded as `string` and a field that need to decoded as `[]string` with one element is appearing on the same way (as a string) on the input json:`{...,"Field:"data",...}`.
## temporary fix:
In order to temporary overcome this issue we added an helper type called `json.Set` type to distinguish between the two types when decoding the string.
```go
type Set []string
```
## what we need to do:
In the future we will make the decoder work on the `[]string` type fields and removing the `json.Set` helper type.

PR #17, opened 2021-03-07
convert all the []string to json.Set on json types

OmerBenHayun:
convert all the `[]string` to `json.Set` on json types as a temporary fix to the decoder problem as described on issue #16 .

### DIFF
--- a/pkg/codegenerator/generator.go
+++ b/pkg/codegenerator/generator.go
@@ -248,7 +248,13 @@ func createFiledLine(fieldName string, filedType interface{}, field string, max 
 	if isSet {
 		maxStr := fmt.Sprintf("%v", max)
 		if maxStr != "1" {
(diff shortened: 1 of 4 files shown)


hunchback:
@OmerBenHayun ci is failing as you did not generate code

pkg/codegenerator/generator.go
```diff
@@ -248,7 +248,13 @@ func createFiledLine(fieldName string, filedType interface{}, field string, max
 	if isSet {
 		maxStr := fmt.Sprintf("%v", max)
 		if maxStr != "1" {
-			return fmt.Sprintf(" %s \t[]%s `json:\"%s,omitempty\"`\n", fieldName, typeConvert(filedType), field)
+			var convertedType string
+			if filedType == "string" {
+				convertedType = "json.Set"
```
roytman, line 253: It is not so simple, Set can contain other types, not only strings.

roytman:
we have moved to use libovsdb library.

PR #22, opened 2021-03-08
fix path to the schemas location

roytman:
Fix the new schemas location in `server.go` 


OmerBenHayun:
lgtm

PR #24, opened 2021-03-08
Ovsdb test

hunchback:


pkg/ovsdb/dbserver.go
```diff
@@ -19,14 +19,25 @@ import (
 	"github.com/ibm/ovsdb-etcd/pkg/types/_Server"
 )
 
+type DBServerInterface interface {
```
roytman, line 22: I would suggest separating OVSDB methods, such as `Lock`, `Unlock`, GetSchema`, GetUUID` and others

pkg/ovsdb/dbserver.go
```diff
@@ -259,3 +270,51 @@ func (con *DBServer) GetMarshaled(prefix string, columns []interface{}) (*[]map[
...
+}
+
+func (con *DBServer) GetUUID() string {
+	return con.uuid
+}
+
+type DBServerMock struct {
+}
+
+func NewDBServerMock() (DBServerInterface, error) {
```
roytman, line 285: I'd prefer to put the Mock into a separate file

pkg/ovsdb/dbserver.go
```diff
@@ -19,14 +19,25 @@ import (
 	"github.com/ibm/ovsdb-etcd/pkg/types/_Server"
 )
 
+type DBServerInterface interface {
+	Lock(ctx context.Context, id string) (bool, error)
+	Unlock(ctx context.Context, id string) error
+	AddSchema(schemaName, schemaFile string) error
```
roytman, line 25: AddSchema and especially LoadServerData are part of the development scaffolding, they will be removed.

pkg/ovsdb/dbserver.go
```diff
@@ -19,14 +19,25 @@ import (
...
+	LoadServerData() error
+	GetData(prefix string, keysOnly bool) (*clientv3.GetResponse, error)
+	GetMarshaled(prefix string, columns []interface{}) (*[]map[string]string, error)
```
roytman, line 28: GetMarshaled, will probably  be removed too

roytman:
we will have to add notifications, the process that starts as a result of etcd events

pkg/ovsdb/dbserver.go
hunchback: I rather this be done in future, so now I just wanted to make the *existing* code testable

pkg/ovsdb/dbserver.go
hunchback: this very C++ or Java but is not the norm in go, as is creates many many very small files (each with overhead of header includes etc.), ... I can show many examples to support this claim in leading go projects

pkg/ovsdb/dbserver.go
hunchback: but this is not related to the focus of this PR, this is future work which should be done but should not be mixed with this effort

pkg/ovsdb/dbserver.go
hunchback: same as above

PR #39, opened 2021-03-14
Adds tests for initial data returned by the monitor functions

roytman:
Adds tests for initial data returned by the monitor functions

### DIFF
--- a/pkg/ovsdb/service_test.go
+++ b/pkg/ovsdb/service_test.go
@@ -5,14 +5,16 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/ibm/ovsdb-etcd/pkg/common"
-
 	"github.com/stretchr/testify/assert"
+
+	"github.com/ibm/ovsdb-etcd/pkg/common"
 )
 
-const baseTransact string = "../../tests/data/transact/"
+const baseData string = "../../tests/data/"
+const baseTransact string = baseData + "/transact/"
 
 func TestTransactSelect(t *testing.T) {
+
 	byteValue, err := common.ReadFile(baseTransact + "select-response.json")
 	assert.Nil(t, err)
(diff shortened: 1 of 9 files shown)


pkg/ovsdb/monitorining_test.go
```diff
@@ -0,0 +1,90 @@
+package ovsdb
```
hunchback, line 1: you may want to consider renaming file to `monitoring_test.go`

pkg/ovsdb/service_test.go
```diff
@@ -5,23 +5,25 @@ import (
...
 
-	"github.com/ibm/ovsdb-etcd/pkg/common"
-
 	"github.com/stretchr/testify/assert"
+
+	"github.com/ibm/ovsdb-etcd/pkg/common"
 )
 
-const baseTransact string = "../../tests/data/transact/"
+const BaseData string = "../../tests/data/"
```
hunchback, line 13: this is only going to be used internally ... anyone using `BaseData` from a different location will fail

pkg/ovsdb/service_test.go
```diff
@@ -5,23 +5,25 @@ import (
...
+const BaseTransact string = BaseData + "/transact/"
```
hunchback, line 14: same as above

pkg/ovsdb/monitorining_test.go
roytman: done

pkg/ovsdb/service_test.go
roytman: ok

pkg/ovsdb/service_test.go
roytman: ok

PR #47, opened 2021-03-16
Monitors sceleton

roytman:


pkg/ovsdb/database.go
```diff
@@ -5,20 +5,23 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/ibm/ovsdb-etcd/pkg/ovsjson"
```
hunchback, line 8: move this down into a block of includes below the standard ones

PR #57, opened 2021-04-18
DRAFT: New txn

hunchback:


pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	if err != nil {
+		txn.response.Error = err.Error()
+		return err
+	}
+
+	return nil
+}
+
+const DELIM = "/"
+const UUID = "_uuid"
```
roytman, line 127: I pretty sure, we have this constant at some place

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	FN_EQ = "=="
+	FN_NE = "!="
+	FN_GE = ">="
+	FN_GT = ">"
+	FN_IN = "includes"
+	FN_EX = "excludes"
+)
+
+func isSupportedFunc(function string) error {
+	if function == FN_EQ || function == FN_NE {
```
roytman, line 161: should we check only these 2 ?

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	if fn == FN_LT && actual < expected {
+		return true, nil
+	}
+	if fn == FN_LE && actual <= expected {
+		return true, nil
+	}
+	return false, nil
+}
+
+func condReal(act interface{}, fn string, exp interface{}) (bool, error) {
```
roytman, line 222: can we combine condInteger and condReal ?

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+		return true, nil
+	}
+	return false, nil
+}
+
+func condTypeSwitch(actual interface{}, fn string, expected interface{}) (bool, error) {
+	switch actual.(type) {
+	case int:
+		return condInteger(actual, fn, expected)
+	case float64:
```
roytman, line 360: float32 ?

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+}
+
+func etcdPutRow(txn *Transaction, table, uuid string, row map[string]interface{}) error {
+	key := makeKey(txn.dbname, table, uuid)
+	setRowUUID(row, uuid)
+	val, err := makeValue(row)
+	if err != nil {
+		return err
+	}
+	etcdOp := clientv3.OpPut(key, val)
```
roytman, line 470: you cannot do it outside of transaction

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	id := "1" // FIXME: missing ovsOp.Id
+	key := makeKey(txn.dbname, "_lock", id)
+	etcdOp := clientv3.OpGet(key)
+	txn.etcdThen = append(txn.etcdThen, etcdOp)
+	return nil
+}
+
+func doAssert(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	id := "1" // FIXME: missing ovsOp.Id
+	_, ok := txn.etcdCache["_lock"][id]
```
roytman, line 704: it's wrong, the lock assert should be part of the etcd transaction condition

roytman (changes requested):

pkg/ovsdb/handler.go
```diff
@@ -35,68 +34,7 @@ type Handler struct {
 }
 
 func (ch *Handler) Transact(ctx context.Context, param []interface{}) (interface{}, error) {
-	klog.V(5).Infof("Transact request, parameters %v", param)
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	OP_MUTATE  = "mutate"
+	OP_DELETE  = "delete"
+	OP_WAIT    = "wait"
+	OP_COMMIT  = "commit"
+	OP_ABORT   = "abort"
+	OP_COMMENT = "comment"
+	OP_ASSERT  = "assert"
+)
+
+func (txn *Transaction) etcdCommit() (*clientv3.TxnResponse, error) {
```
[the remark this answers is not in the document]
hunchback: it does the commit so I rather keep this name

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	etcdCache     map[string]map[string]map[string]interface{}
+}
+
+func NewTransaction(dbname string, operations []libovsdb.Operation, endpoints []string) *Transaction {
+	txn := new(Transaction)
+	txn.dbname = dbname
+	txn.operations = operations
+	txn.response.Result = make([]libovsdb.OperationResult, len(operations))
+	txn.etcdCtx = context.TODO()
+	txn.etcdEndpoints = endpoints
```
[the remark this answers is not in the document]
hunchback: I keep the endpoints so that I can do the Open at a later stage then during creation of the object

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	return txn
+}
+
+func (txn *Transaction) Open() error {
+	cli, err := clientv3.New(clientv3.Config{Endpoints: txn.etcdEndpoints})
```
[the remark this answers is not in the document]
hunchback: no, for now doing i rather its' independent (better for my testing again etcd)

pkg/ovsdb/transact.go
hunchback: agree, I have removed this and now we check everything according to spec

pkg/ovsdb/transact.go
hunchback: no. as we depend on type conversion so while code looks the same one works on int the other on float64 (this is the compromise when using a strongly typed lang such as go)

pkg/ovsdb/transact.go
hunchback: no. this is how Real is interpreted (I think I saw this in libovsdb that real is float64)

pkg/ovsdb/transact.go
hunchback: as explained this is not the transaction only preparation for it

pkg/ovsdb/transact.go
hunchback: will address this later

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,918 @@
...
+		if !ok {
+			continue
+		}
+		ovsResult.Count = ovsResult.Count + 1
+		err = RowMutate(&row, ovsOp.Mutations)
+		if err != nil {
+			return err
+		}
+		txn.etcdCache[ovsOp.Table][uuid] = row
+		etcdPutRow(txn, ovsOp.Table, uuid, row)
```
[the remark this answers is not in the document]
hunchback: yes ... in one of the next iterations

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,918 @@
...
+		}
+	}
+	ovsResult.Count = ovsResult.Count + 1
+	uuid := uuid.New() /* generate RFC4122 UUID */
+	return etcdPutRow(txn, ovsOp.Table, uuid.String(), ovsOp.Row)
+}
+
+/* select */
+func preSelect(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	ok, err := doesWhereContainCondTypeUUID(ovsOp.Where)
```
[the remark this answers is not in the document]
hunchback: I don't allow to intermix select with other operations, thus I now fetch the data in `pre` and process it in `do`

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,918 @@
...
+	}
+	if ok {
+		return nil
+	}
+	etcdGetTable(txn, ovsOp.Table)
+	return nil
+}
+
+func doSelect(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	for _, row := range txn.etcdCache[ovsOp.Table] {
```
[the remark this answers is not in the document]
hunchback: I always work via cache, this way if any other operations come before the current one then the current operation will be working on the up-to-date value not a stale value, for example think 2 mutate operations which do a `+=1` operation

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,918 @@
...
+	txn.etcdThen = append(txn.etcdThen, etcdOp)
+	return nil
+}
+
+/* insert */
+func preInsert(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	if ovsOp.UUIDName == "" {
+		return nil
+	}
+	etcdGetTable(txn, ovsOp.Table)
```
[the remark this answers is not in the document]
hunchback: please explain

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,918 @@
...
+	etcdOp := clientv3.OpPut(key, comment)
+	txn.etcdThen = append(txn.etcdThen, etcdOp)
+	return nil
+}
+
+/* assert */
+func preAssert(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	id := "1" // FIXME: missing ovsOp.Id
+	key := makeKey(txn.dbname, "_lock", id)
+	etcdOp := clientv3.OpGet(key)
```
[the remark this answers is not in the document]
hunchback: deleted the implementation for now, will get back to it later

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,918 @@
...
+	}
+	if ok {
+		return nil
+	}
+	etcdGetTable(txn, ovsOp.Table)
+	return nil
+}
+
+func doDelete(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	for uuid, row := range txn.etcdCache[ovsOp.Table] {
```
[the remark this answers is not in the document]
hunchback: I will need to warm the cache, so to keep uniformity (I want to avoid multiple paths of execution, will optimize and create different paths only if proven that this causes a performance issue which is tangible)

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	if !ok {
+		return false, errors.New(E_CONSTRAINT_VIOLATION)
+	}
+	fn := c.Function
+	expected, ok := c.Value.(libovsdb.UUID)
+	if !ok {
+		return false, errors.New(E_CONSTRAINT_VIOLATION)
+	}
+
+	if (fn == FN_EQ || fn == FN_IN) && actual == expected {
```
roytman, line 439: UUID is a struct, I'm not sure that "==" will correctly work. Should we use deep copy, or their "GoUUID" values

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+		return "", errors.New(E_ERROR)
+	}
+	condition, err := NewCondition(cond2)
+	if err != nil {
+		return "", err
+	}
+	if condition.Column != COL_UUID {
+		return "", nil
+	}
+	if condition.Function != FN_EQ {
```
roytman, line 521: + FN_IN

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+		return "", nil
+	}
+	ovsUUID, ok := condition.Value.(libovsdb.UUID)
+	if !ok {
+		return "", errors.New(E_ERROR)
+	}
+	_, err = guuid.Parse(ovsUUID.GoUUID)
```
roytman, line 528: can we reuse the logic from uuid.validateUUID()

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+
+func isRowSelectedByCond(row *map[string]interface{}, cond []interface{}) (bool, error) {
+	condition, err := NewCondition(cond)
+	if err != nil {
+		return false, err
+	}
+	return condition.Compare(row)
+}
+
+func reduceRowByColumns(row *map[string]interface{}, columns []string) *map[string]interface{} {
```
roytman, line 578: I have similar functionality in Monitors

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	}
+}
+
+func etcdGetByWhere(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	uuid, err := doesWhereContainCondTypeUUID(ovsOp.Where)
+	if err != nil {
+		ovsResult.Error = err.Error()
+		return err
+	}
+	etcdGetTableOrRow(txn, ovsOp.Table, uuid)
```
roytman, line 851: if you have 2 operations on the same table, will you bring it twice ?

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	txn.etcdThen = append(txn.etcdThen, etcdOp)
+	return nil
+}
+
+/* insert */
+func preInsert(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	if ovsOp.UUIDName == "" {
+		return nil
+	}
+	etcdGetTable(txn, ovsOp.Table)
```
roytman, line 892: In order to validate uniqueness of a new uuid (I assume that you fetch the table for that), you don't need to fetch entire table, enough to bring only keys. (etc has this option)

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	if ovsOp.UUIDName != "" {
+		for _, row := range txn.data.Table(ovsOp.Table) {
+			if (*row)[COL_UUIDNAME] == ovsOp.UUIDName {
+				err := errors.New(E_DUP_UUIDNAME)
+				ovsResult.Error = err.Error()
+				return err
+			}
+		}
+	}
+	ovsResult.Count = ovsResult.Count + 1
```
roytman, line 906: it should be taken from the etcd response

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+		ovsResult.Rows = append(ovsResult.Rows, *resultRow)
+	}
+	return nil
+}
+
+/* update */
+func preUpdate(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	return etcdGetByWhere(txn, ovsOp, ovsResult)
+}
+
```
roytman, line 938: TODOD: add version check

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+		etcdPutRow(txn, ovsOp.Table, uuid, row)
+	}
+	return nil
+}
+
+/* mutate */
+func preMutate(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	return etcdGetByWhere(txn, ovsOp, ovsResult)
+}
+
```
roytman, line 963: TODO: add version check

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+func doMutate(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	for uuid, row := range txn.data.Table(ovsOp.Table) {
+		ok, err := isRowSelectedByWhere(row, ovsOp.Where)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			continue
+		}
+		ovsResult.Count = ovsResult.Count + 1
```
roytman, line 973: result from etcd

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+		}
+		*(txn.data.Row(ovsOp.Table, uuid)) = *row
+		etcdPutRow(txn, ovsOp.Table, uuid, row)
+	}
+	return nil
+}
+
+/* delete */
+func preDelete(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	return etcdGetByWhere(txn, ovsOp, ovsResult)
```
roytman, line 986: if you have uuid, you don't need fetch

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+}
+
+/* comment */
+func preComment(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	return nil
+}
+
+func doComment(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
+	now := time.Now()
+	key := makeKey(txn.request.DBName, "_comment", now.String())
```
roytman, line 1080: format according OVSDB

roytman (commented):
`transact_test.go` will check tomorrow

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+				Op:    OP_INSERT,
+				Table: TABLE1,
+				Row: map[string]interface{}{
+					"key1": "val1",
+				},
+			},
+		},
+	}
+	testEtcdCleanup(t, DBNAME1, TABLE1)
+	resp := testTransact(t, req)
```
roytman, line 110: you don't test corner cases, e.g. when uuid exists

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+	assert.Equal(t, "", resp.Error)
+	dump := testEtcdDump(t, DBNAME1, TABLE1)
+	assert.Equal(t, "val1", dump["key1"])
+}
+
+func TestTransactSelect(t *testing.T) {
+	req := &libovsdb.Transact{
+		DBName: DBNAME1,
+		Operations: []libovsdb.Operation{
+			{
```
roytman, line 120: the test case should be extended to case when the Select operation contains Columns and Where

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	testEtcdCleanup(t, DBNAME1, TABLE1)
+	testEtcdPut(t, DBNAME1, TABLE1, "key1", "val1")
+	resp := testTransact(t, req)
+	assert.Equal(t, "", resp.Error)
+	dump := testEtcdDump(t, DBNAME1, TABLE1)
+	assert.Equal(t, "val1", dump["key1"])
+}
+
+func TestTransactUpdate(t *testing.T) {
```
roytman, line 134: maybe to rename to TestTransactionInsert, according to its operation name

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+	req := &libovsdb.Transact{
+		DBName: DBNAME1,
+		Operations: []libovsdb.Operation{
+			{
+				Op:    OP_INSERT,
+				Table: TABLE1,
```
roytman, line 140: you do 2 operations on the same key during the same transaction. Make sane to add a separate test with 2 independent operations.

We need unit tests with Where == uuid, this is very common use case. (Relevant for other tests as well)

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+	assert.Equal(t, "", resp.Error)
+	dump := testEtcdDump(t, DBNAME1, TABLE1)
+	assert.Equal(t, "val2", dump["key1"])
+}
+
+func TestTransactMutate(t *testing.T) {
+	req := &libovsdb.Transact{
+		DBName: DBNAME1,
+		Operations: []libovsdb.Operation{
+			{
```
roytman, line 166: the same comments as for Insert

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+				Table: TABLE1,
+				Row: map[string]interface{}{
+					"key1": "val1",
+				},
+			},
+		},
+	}
+	testEtcdCleanup(t, DBNAME1, TABLE1)
+	resp := testTransact(t, req)
+	assert.Equal(t, "", resp.Error)
```
roytman, line 111: in the correct case the error should be nil, and not ""

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+		Operations: []libovsdb.Operation{
+			{
+				Op:    OP_SELECT,
+				Table: TABLE1,
+			},
+		},
+	}
+	testEtcdCleanup(t, DBNAME1, TABLE1)
+	testEtcdPut(t, DBNAME1, TABLE1, "key1", "val1")
+	resp := testTransact(t, req)
```
roytman, line 128: I didn't see and checks of transaction response, you check only error, what about the result objects

pkg/libovsdb/notation.go
```diff
@@ -14,6 +17,8 @@ type Operation struct {
 	Where     []interface{}            `json:"where,omitempty"`
 	Until     string                   `json:"until,omitempty"`
 	UUIDName  string                   `json:"uuid-name,omitempty"`
+	Comment   string                   `json:"comment,omitempty"`
+	Durable   bool                     `json:"durable,omitempty"`
```
roytman, line 21: according to the OVSDB extensions:
https://docs.openvswitch.org/en/latest/ref/ovsdb-server.7/#insert

> As an extension, Open vSwitch 2.13 and later allow an optional uuid member to specify the UUID for the new row. The specified UUID must be unique within the table when it is inserted and not the UUID of a row previously deleted within the transaction. If the UUID violates these rules, then the operation fails with a duplicate uuid error.

pkg/ovsdb/transact.go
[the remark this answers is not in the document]
hunchback: see comment I placed in the code

Makefile
```diff
@@ -38,7 +38,8 @@ verify: $(VERIFY)
 
 .PHONY: etcd
 etcd:
-	$(MAKE) -C tests/e2e/ etcd &
+	sudo systemctl restart etcd
+	sudo systemctl restart etcd
```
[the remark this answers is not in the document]
hunchback: mistake should be status .. will fix

pkg/cmd/server/server.go
```diff
@@ -43,11 +43,14 @@ func main() {
 		klog.Fatal("Wrong ETCD members list", etcdMembers)
 	}
 	etcdServers := strings.Split(*etcdMembers, ",")
-	db, err := ovsdb.NewDatabaseEtcd(etcdServers, *databasePrefix)
+
+	cli, err := ovsdb.NewEtcdClient(etcdServers)
```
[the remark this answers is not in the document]
hunchback: as the database abstraction basically implements: monitoring, transaction, locks, etcd these are mostly independent (other than sharing the client connection and agreeing on path format) so I see no reason to bundle them all together, on the contrary. So rather than the transaction fighting the database interface to to etcd operations (and use clientv3 data structures) these are all used directly.

To bridge the fact that currently code was meant to use the database abstraction is to extract the small bits of functionality which are shared namely:
- creation of etcd client
- generation of etcd key (from service, database, table, uuid)
- generation of etcd value (json serialization of row)

For now I have extract the creation of the etcd client. So you can see that the creation of the client has been made external to the database and it now get the client and does not have to create the client itself (from the endpoints)

pkg/libovsdb/notation.go
```diff
@@ -101,12 +106,60 @@ func NewMutation(column string, mutator string, value interface{}) []interface{}
...
+	Operations []Operation
+}
+
+func NewTransact(params []interface{}) (*Transact, error) {
+	if len(params) < 2 {
+		return nil, errors.New("malformed transaction")
+	}
+
+	tx := new(Transact)
+	for i, v := range params {
```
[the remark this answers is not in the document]
hunchback: how?

pkg/libovsdb/notation.go
```diff
@@ -101,12 +106,60 @@ func NewMutation(column string, mutator string, value interface{}) []interface{}
...
+	return tx, nil
+}
+
 // TransactResponse represents the response to a Transact Operation
 type TransactResponse struct {
 	Result []OperationResult `json:"result"`
 	Error  string            `json:"error"`
 }
 
+func (tx *TransactResponse) ReturnValue() (string, error) {
```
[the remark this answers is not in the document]
hunchback: it was crashing when I did that ... let me debug and re-check

pkg/ovsdb/transact.go
hunchback: will do (see start of transact.go file)

pkg/ovsdb/transact.go
hunchback: agree (at start of file)

pkg/ovsdb/transact.go
hunchback: this result is as good as any. If the transaction was a success then this number of accurate

pkg/ovsdb/transact.go
hunchback: as explain, creating different paths of execution (with added code complexity, testing, etc.) needs justification. the general justification of "better performance" is counter-productive as it will force optimized code (and complexity) in places these are not needed. Currently focusing on correctness and simplicity it think its wrong to bring-in optimizations here (and elsewhere in the code) and this preliminary stage.

pkg/ovsdb/transact.go
hunchback: agree ... fixing

pkg/libovsdb/notation.go
```diff
@@ -101,12 +106,60 @@ func NewMutation(column string, mutator string, value interface{}) []interface{}
...
+		}
+	}
+
+	return tx, nil
+}
+
 // TransactResponse represents the response to a Transact Operation
 type TransactResponse struct {
 	Result []OperationResult `json:"result"`
 	Error  string            `json:"error"`
```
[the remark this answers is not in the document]
hunchback: so how does one know that the operation failed ?

pkg/ovsdb/transact.go
roytman: >  If the transaction was a success then this number is accurate

It is not always true. for example delete row. Another point is a demonstration of which the operations failed.

pkg/libovsdb/notation.go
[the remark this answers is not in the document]
roytman: yes, the transaction object has a lot of other field, originally I thought about 
Transaction struct {
     dbName sting  
     operations []Operations
}

pkg/libovsdb/notation.go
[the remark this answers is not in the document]
roytman: From the previous spec.
... and a "result" member that is an array with the same number of elements as "params".  Each element of the "result" array corresponds to the same element of the "params" array.  The "result" array elements may be
   interpreted as follows:

   o  A JSON object that does not contain an "error" member indicates
      that the operation completed successfully.  The specific members
      of the object are specified below in the descriptions of
      individual operations.  Some operations do not produce any
      results, in which case the object will have no members.
   o  An <error> indicates that the matching operation completed with an
      error.
   o  A JSON null value indicates that the operation was not attempted
      because a prior operation failed.

pkg/ovsdb/transact.go
roytman: in addition, see my comment about the transaction response

pkg/ovsdb/transact.go
roytman: we can check the OVSDB deployments

pkg/ovsdb/transact.go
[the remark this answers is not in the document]
hunchback: s/etcdCommit/etcdTranaction/

pkg/ovsdb/transact.go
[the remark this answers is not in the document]
hunchback: need to remove

pkg/libovsdb/notation.go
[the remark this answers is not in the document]
hunchback: ok

pkg/libovsdb/notation.go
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
+package ovsdb
+
+// TODO: named-uuid from one operation servces other operation
+// TODO: add protection (cond) for transactions
+// TODO: add re-try mechanism for transactions
+// TODO: add schema to initiate default values
+// TODO: add schema to trigger correct type based mutate
+// TODO: add schema and verify no attempt to mutate/update a read-only column
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	E_DUP_UUIDNAME         = "duplicate uuid-name"
+	E_CONSTRAINT_VIOLATION = "constraint violation"
+	E_DOMAIN_ERROR         = "domain error"
+	E_RANGE_ERROR          = "range error"
+	E_TIMEOUT              = "timed out"
+	E_NOT_SUPPORTED        = "not supported"
+	E_ABORTED              = "aborted"
+	E_NOT_OWNER            = "not owner"
+	/* extra errors */
+	E_ERROR = "error"
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+)
+
+func isEqualRows(a, b []map[string]interface{}) bool {
+	return reflect.DeepEqual(a, b)
+}
+
+func isEqualSet(a, b libovsdb.OvsSet) bool {
```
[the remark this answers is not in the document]
hunchback: will move later

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	return reflect.DeepEqual(a.GoSet, b.GoSet) // XXX: should I sort first?
+}
+
+func isEqualMap(a, b libovsdb.OvsMap) bool {
```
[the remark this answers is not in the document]
hunchback: will move later

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	return reflect.DeepEqual(a.GoMap, b.GoMap)
+}
+
+func isEqual(a, b interface{}) bool {
+	return reflect.DeepEqual(a, b)
+}
+
+const DELIM = "/"
+
+const (
```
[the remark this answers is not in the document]
hunchback: agree

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	OP_MUTATE  = "mutate"
+	OP_DELETE  = "delete"
+	OP_WAIT    = "wait"
+	OP_COMMIT  = "commit"
+	OP_ABORT   = "abort"
+	OP_COMMENT = "comment"
+	OP_ASSERT  = "assert"
+)
+
+func NewEtcdClient(endpoints []string) (*clientv3.Client, error) {
```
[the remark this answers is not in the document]
hunchback: moved to db

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	}
+
+	txn.etcdIf = []clientv3.Cmp{}
+	txn.etcdThen = []clientv3.Op{}
+	txn.etcdElse = []clientv3.Op{}
+
+	return res, err
+}
+
+type KeyValue struct {
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+
+type Cache struct {
+	db Database
+}
+
+func (c *Cache) Table(table string) Table {
+	if c.db == nil {
+		c.db = Database{}
+	}
+	t, ok := c.db[table]
```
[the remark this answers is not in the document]
hunchback: per transaction (for now)

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	txn.request = *request
+	txn.response.Result = make([]libovsdb.OperationResult, len(request.Operations))
+	txn.etcdCtx = context.TODO()
+	txn.etcdCli = cli
+	return txn
+}
+
+type ovsOpCallback func(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error
+
+var ovsOpCallbackMap = map[string][]ovsOpCallback{
```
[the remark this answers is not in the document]
hunchback: yes

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	OP_COMMENT: {preComment, doComment},
+	OP_ASSERT:  {preAssert, doAssert},
+}
+
+func (txn *Transaction) Commit() error {
+
+	var err error
+
+	/* verify that select is not intermixed with other operations */
+	hasSelect := false
```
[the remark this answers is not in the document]
hunchback: for now we assume that select is not inter-mixed ... if we find a use case then we will change

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,711 @@
...
+	var err error
+
+	/* fetch needed data from database needed to perform the operation */
+	for i, ovsOp := range txn.operations {
+		if err = ovsOpCallbackMap[ovsOp.Op][0](txn, &ovsOp, &txn.response.Result[i]); err != nil {
+			txn.response.Error = err.Error()
+			return err
+		}
+	}
+	_, err = txn.etcdCommit()
```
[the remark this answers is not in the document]
hunchback: one for pre and one for do

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	if err != nil {
+		txn.response.Error = err.Error()
+		return err
+	}
+
+	return nil
+}
+
+func makePrefix(dbname, table string) string {
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	return dbname + DELIM + table
+}
+
+func makeKey(dbname, table, uuid string) string {
+	return dbname + DELIM + table + DELIM + uuid
+}
+
+func makeValue(row *map[string]interface{}) (string, error) {
```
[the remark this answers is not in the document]
hunchback: move to (db? row? utils? TBD)

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	b, err := json.Marshal(*row)
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
+
+func setRowUUID(row *map[string]interface{}, uuid string) {
```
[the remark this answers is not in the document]
hunchback: agree

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+}
+
+func NewCondition(condition []interface{}) (*Condition, error) {
+	if len(condition) != 3 {
+		return nil, errors.New(E_ERROR)
+	}
+
+	column, ok := condition[0].(string)
+	if !ok {
+		return nil, errors.New(E_ERROR)
```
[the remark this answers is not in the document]
hunchback: added a TODO

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	}
+
+	fn, ok := condition[1].(string)
+	if !ok {
+		return nil, errors.New(E_ERROR)
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
```diff
@@ -0,0 +1,1094 @@
...
+	}
+
+	value := condition[2]
+	return &Condition{Column: column, Function: fn, Value: value}, nil
+}
+
+func (c *Condition) CompareInteger(row *map[string]interface{}) (bool, error) {
+	actual, ok := (*row)[c.Column].(int)
+	if !ok {
+		return false, errors.New(E_CONSTRAINT_VIOLATION)
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: currently yes, should be merged into one call

pkg/ovsdb/transact.go
hunchback: delayed for when we address optimizations

pkg/ovsdb/transact.go
hunchback: if we can ensure the database did not change then this result is accurate, if not then we need to use the value calculated from the etcd response (if we optimize delete then this may be relevant).

pkg/ovsdb/transact_test.go
```diff
@@ -0,0 +1,268 @@
...
+func testEtcdCleanupComment(t *testing.T, dbname string) {
+	testEtcdCleanup(t, dbname, "_comment")
+}
+
+func testMergeKvs(kvs []*mvccpb.KeyValue, table string) (map[string]interface{}, error) {
+	dump := map[string]interface{}{}
+	for _, x := range kvs {
+		kv, err := NewKeyValue(x)
+		if err != nil {
+			return dump, err
```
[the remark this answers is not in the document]
hunchback: ok

pkg/ovsdb/transact_test.go
hunchback: for later

pkg/ovsdb/transact_test.go
hunchback: ok

pkg/ovsdb/transact_test.go
hunchback: later

pkg/ovsdb/transact_test.go
hunchback: later

pkg/ovsdb/transact_test.go
hunchback: it checks update (insert is just for the being able to test the update)

pkg/ovsdb/transact_test.go
hunchback: later

pkg/ovsdb/transact_test.go
hunchback: later

hunchback:
@roytman i am leaving prototype of `testEtcdDump()` as existed.

roytman:
@hunchback , what about a possible uuid filed as part of the transaction object? Or, another TODO (which is even better) I concern that we'd forget about it.

PR #61, opened 2021-04-20
Add single emelemnt set support, fix maps compare

roytman:

### DIFF
--- a/pkg/libovsdb/encoding_test.go
+++ b/pkg/libovsdb/encoding_test.go
@@ -2,48 +2,162 @@ package libovsdb
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
-// empty Set test
-func TestEmptySet(t *testing.T) {
-	emptySet, err := NewOvsSet([]string{})
-	assert.Nil(t, err)
-	jsonStr, err := json.Marshal(emptySet)
-	assert.Nil(t, err)
-	expected := "[\"set\",[]]"
-	assert.JSONEqf(t, expected, string(jsonStr), "they should be equal\n")
+type marshalSetTestTuple struct {
+	objInput           interface{}
+	jsonExpectedOutput string
 }
 
-// test Set
-func TestSet(t *testing.T) {
-	ovsSet, err := NewOvsSet([]string{"aa", "bb"})
-	assert.Nil(t, err)
-	jsonStr, err := json.Marshal(ovsSet)
-	assert.Nil(t, err)
-	expected := "[\"set\",[\"aa\",\"bb\"]]"
-	require.JSONEqf(t, expected, string(jsonStr), "they should be equal\n")
+type marshalMapsTestTuple struct {
+	objInput           map[string]string
+	jsonExpectedOutput string
 }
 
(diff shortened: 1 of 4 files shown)


pkg/libovsdb/encoding_test.go
```diff
@@ -2,48 +2,214 @@ package libovsdb
...
+	for _, elem := range set2.GoSet {
+		switch elem.(type) {
+		case UUID:
+			uuid := elem.(UUID)
+			res2[uuid.GoUUID] = true
+		default:
+			s := fmt.Sprintf("%v", elem)
+			res2[s] = true
+		}
+	}
```
hunchback, line 151: this is a total code duplication of the block above, please use functions ...

pkg/libovsdb/encoding_test.go
```diff
@@ -2,48 +2,214 @@ package libovsdb
...
+		jsonExpectedOutput: `"aa"`,
+	},
+	{
+		objInput:           []string{`aa`, `bb`},
+		jsonExpectedOutput: `["set",["aa","bb"]]`,
+	},
+	{
+		objInput:           [2]string{`aa`, `bb`},
+		jsonExpectedOutput: `["set",["aa","bb"]]`,
+	},
```
hunchback, line 62: redundant, this is exactly as above (after go compilation its' identical).

pkg/libovsdb/encoding_test.go
```diff
@@ -2,48 +2,214 @@ package libovsdb
...
+	{
+		objInput:           []int{10, 15},
+		jsonExpectedOutput: `["set",[10,15]]`,
+	},
+	{
+		objInput:           []uint16{10, 15},
+		jsonExpectedOutput: `["set",[10,15]]`,
+	},
```
hunchback, line 70: not sure why you are testing all simple types? if you test for string then no need to test for int, int32, float32, etc, etc. this adds no value as the focus here is to the OvsSet (which can contain anything) not serialization of all simple tests (something that is handled by standard json encoder/decoder).

hunchback:
@roytman some higher level comments:

(1)
There is a lot of redundancy (for example no need to test for all types of set elements just use one simple type such as string).

(2)
There is code duplication - use helper functions even if this is test code

(3)
You separated Mashal and Unmarshal but combining both into a single test makes a lot of sense and enables to test both.

(4)
You need to have a standard (sorted) representation in json of OvsSet and ovsMap once this is the case all needed is to perform a string comparison. By doing this you eliminate most of the code complexity.

pkg/libovsdb/encoding_test.go
roytman: No, they are not, one case is slice and another is array

pkg/libovsdb/encoding_test.go
roytman: the function was removed

roytman:
@hunchback I think I resolved all your comments, most of the set tests are to validate corner cases such as single element, array of elements, uuid, named-uuid ....
We don't need a standard (sorted) representation in json of OvsSet and ovsMap, there is `assert.ElementsMatch` instead.

pkg/libovsdb/encoding_test.go
hunchback: ok

PR #63, opened 2021-04-25
[WIP] Global monitors

roytman:

### DIFF
--- a/pkg/ovsdb/database.go
+++ b/pkg/ovsdb/database.go
@@ -12,27 +12,28 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/concurrency"
 	"k8s.io/klog/v2"
-
-	"github.com/ibm/ovsdb-etcd/pkg/ovsjson"
 )
 
 type Databaser interface {
 	GetLock(ctx context.Context, id string) (Locker, error)
+	AddMonitors(dbName string, updaters map[string][]*updater, handler handlerKey)
+	RemoveMonitors(dbName string, updaters map[string][]string, handler handlerKey)
+	RemoveMonitor(dbName string)
 	AddSchema(schemaName, schemaFile string) error
 	GetData(prefix string, keysOnly bool) (*clientv3.GetResponse, error)
 	PutData(ctx context.Context, key string, obj interface{}) error
 	GetSchema(name string) (string, bool)
-	AddMonitor(prefix string, mcr ovsjson.MonitorCondRequest, isV1 bool, hand *handlerKey)
-	DelMonitor(prefix string, mcr ovsjson.MonitorCondRequest, isV1 bool, hand *handlerKey)
 }
 
 type DatabaseEtcd struct {
-	cli     *clientv3.Client
+	cli *clientv3.Client
+	// dataBaseName -> schema
 	Schemas map[string]string
-	// map from prefix to monitors
-	monitors map[string]monitor
-	mu       sync.Mutex
-	prefix   string
+	mu      sync.Mutex
+	prefix  string
+	// databaseName -> monitor
+	// We have a single monitor (etcd watcher) per database
+	monitors map[string]*monitor
 }
 
 type Locker interface {
@@ -71,7 +72,7 @@ func NewDatabaseEtcd(cli *clientv3.Client, prefix string) (Databaser, error) {
 		prefix = prefix + "/"
 	}
 	return &DatabaseEtcd{cli: cli,
-		Schemas: make(map[string]string), prefix: prefix}, nil
+		Schemas: map[string]string{}, prefix: prefix, monitors: map[string]*monitor{}}, nil
 }
 
 func (con *DatabaseEtcd) GetLock(ctx context.Context, id string) (Locker, error) {
@@ -135,32 +136,55 @@ func (con *DatabaseEtcd) PutData(ctx context.Context, key string, obj interface{
 	}
 	return nil
 }
-func (con *DatabaseEtcd) AddMonitor(keysPrefix string, mcr ovsjson.MonitorCondRequest, isV1 bool, hand *handlerKey) {
+
+func (con *DatabaseEtcd) AddMonitors(dbName string, updaters map[string][]*updater, handler handlerKey) {
 	con.mu.Lock()
 	defer con.mu.Unlock()
-	if monitor, ok := con.monitors[keysPrefix]; !ok {
-		con.monitors[keysPrefix] = *newMonitor(con.cli, con.prefix+keysPrefix, mcr, isV1, hand)
-	} else {
-		monitor.addHandler(mcr, isV1, hand)
+	m, ok := con.monitors[dbName]
+	if !ok {
+		// we don't have monitors for this database yet, create a new one.
+		m = newMonitor(dbName, con)
+		ctxt, cancel := context.WithCancel(context.Background())
+		m.cancel = cancel
+		wch := con.cli.Watch(clientv3.WithRequireLeader(ctxt), con.prefix+dbName,
+			clientv3.WithPrefix(),
+			clientv3.WithCreatedNotify(),
+			clientv3.WithPrevKV())
+		m.watchChannel = wch
+		con.monitors[dbName] = m
+	}
+	m.addUpdaters(con.prefix, updaters, handler)
+	if !ok {
+		m.start()
 	}
 }
 
-func (con *DatabaseEtcd) DelMonitor(keysPrefix string, mcr ovsjson.MonitorCondRequest, isV1 bool, hand *handlerKey) {
+func (con *DatabaseEtcd) RemoveMonitors(dbName string, updaters map[string][]string, handler handlerKey) {
 	con.mu.Lock()
 	defer con.mu.Unlock()
-	if monitor, ok := con.monitors[keysPrefix]; !ok {
-		klog.Warningf("Delete unexisting monitor from %s", keysPrefix)
-	} else {
-		if monitor.delHandler(mcr, isV1, hand) {
-			delete(con.monitors, keysPrefix)
-		}
+	m, ok := con.monitors[dbName]
+	if !ok {
+		klog.Warningf("Remove nonexistent db monitor %s", dbName)
+		return
 	}
+	m.removeUpdaters(con.prefix, updaters, handler)
+	if !m.hasHandlers() {
+		m.cancel()
+		delete(con.monitors, dbName)
+	}
+}
+
+func (con *DatabaseEtcd) RemoveMonitor(dbName string) {
+	con.mu.Lock()
+	defer con.mu.Unlock()
+	delete(con.monitors, dbName)
 }
 
 type DatabaseMock struct {
 	Response interface{}
 	Error    error
 	Ok       bool
+	mu       sync.Mutex
 }
 
 type LockerMock struct {
@@ -215,8 +239,11 @@ func (con *DatabaseMock) GetUUID() string {
 	return con.Response.(string)
 }
 
-func (con *DatabaseMock) AddMonitor(keysPrefix string, mcr ovsjson.MonitorCondRequest, isV1 bool, hand *handlerKey) {
+func (con *DatabaseMock) AddMonitors(dbName string, updaters map[string][]*updater, handler handlerKey) {
 }
 
-func (con *DatabaseMock) DelMonitor(keysPrefix string, mcr ovsjson.MonitorCondRequest, isV1 bool, hand *handlerKey) {
+func (con *DatabaseMock) RemoveMonitors(dbName string, updaters map[string][]string, handler handlerKey) {
+}
+
+func (con *DatabaseMock) RemoveMonitor(dbName string) {
 }
--- a/pkg/ovsdb/handler.go
+++ b/pkg/ovsdb/handler.go
@@ -29,10 +29,9 @@ type Handler struct {
 	handlerContext context.Context
 
 	mu sync.Mutex
-	// map from jason-values to monitors
-	jsonValueToMonitors               map[interface{}][]*monitor
-	jsonValueToUpdateNotificationType map[interface{}]ovsjson.UpdateNotificationType
 
+	// jsonValue -> handlerMonitorData
+	monitors      map[interface{}]handlerMonitorData
 	databaseLocks map[string]Locker
 }
 
@@ -55,45 +54,24 @@ func (ch *Handler) Cancel(ctx context.Context, param interface{}) (interface{}, 
 
 func (ch *Handler) Monitor(ctx context.Context, param ovsjson.CondMonitorParameters) (interface{}, error) {
 	klog.V(5).Infof("Monitor request, parameters %v", param)
-	ch.mu.Lock()
-	defer ch.mu.Unlock()
-	if err := ch.createOrDieMonitorsArray(param.JsonValue.(string)); err != nil {
+	if err := ch.monitor(param, ovsjson.Update); err != nil {
 		return nil, err
 	}
-	ch.jsonValueToUpdateNotificationType[param.JsonValue.(string)] = ovsjson.Update
-	monitors := ch.jsonValueToMonitors[param.JsonValue]
-
-	for tableName, mcrs := range param.MonitorCondRequests {
-		// TODO handle Where, if Where contains uuid, it can be a part of the key
-		key := fmt.Sprintf("%s/%s", param.DatabaseName, tableName)
-		for _, mcr := range mcrs {
-			ch.db.AddMonitor(key, mcr, true, &handlerKey{ch, param.JsonValue})
-			// TODO add monitor
-			monitors = append(monitors, nil)
-		}
-		// TODO check if we need it?
-		ch.jsonValueToMonitors[param.JsonValue] = monitors
-	}
-	return ch.getMonitoredData(param.DatabaseName, param.MonitorCondRequests, false)
+	return ch.getMonitoredData(param.DatabaseName, param.MonitorCondRequests, true)
 }
 
 func (ch *Handler) MonitorCancel(ctx context.Context, param interface{}) (interface{}, error) {
 	klog.V(5).Infof("MonitorCancel request, parameters %v", param)
-	jsonValue, err := common.ParamsToString(param)
-	if err != nil {
-		return nil, err
-	}
+	jsonValue := param
+
 	ch.mu.Lock()
 	defer ch.mu.Unlock()
-	monitors, ok := ch.jsonValueToMonitors[jsonValue]
+	monitorHandler, ok := ch.monitors[jsonValue]
 	if !ok {
 		return nil, fmt.Errorf("unknown monitor")
 	}
-	for range monitors {
-		// TODO
-	}
-	delete(ch.jsonValueToMonitors, jsonValue)
-	delete(ch.jsonValueToUpdateNotificationType, jsonValue)
+	ch.db.RemoveMonitors(monitorHandler.dataBaseName, monitorHandler.updaters, handlerKey{handler: ch, jsonValue: jsonValue})
+	delete(ch.monitors, jsonValue)
 	return "{}", nil
 }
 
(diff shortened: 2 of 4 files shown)


hunchback:
@roytman a general comment is that I don't think the database abstraction is good for our use, I think every component which uses etcd should get a `*clientv3.Client` and then use clientv3 directly, I suggest this approuch as our initial implementation will be tightly coupled to etcd, thus making the code simpler (as we don't attempt to provide a generic database interface).

pkg/ovsdb/database.go
```diff
@@ -12,28 +13,28 @@ import (
...
 	"k8s.io/klog/v2"
-
-	"github.com/ibm/ovsdb-etcd/pkg/ovsjson"
 )
 
 type Databaser interface {
 	GetLock(ctx context.Context, id string) (Locker, error)
+	AddMonitors(dbName string, updaters map[string][]*updater, handler handlerKey)
+	RemoveMonitors(dbName string, updaters map[string][]string, handler handlerKey)
+    RemoveMonitor(dbName string)
```
hunchback, line 22: following my comment (that we should not use the database abstraction) it would be better to call clientv3 directly from monitor implementation

pkg/ovsdb/database.go
```diff
@@ -82,13 +83,21 @@ func NewDatabaseEtcd(endpoints []string, prefix string) (Databaser, error) {
...
 	return &DatabaseEtcd{cli: cli,
-		Schemas: make(map[string]string), prefix: prefix}, nil
+		Schemas: map[string]string{}, prefix: prefix, monitors: map[string]monitor{} }, nil
 }
 
 func (con *DatabaseEtcd) Close() {
 	con.cli.Close()
 }
 
+func (con *DatabaseEtcd) AddWatchHandler(ctx context.Context, prefix string ) clientv3.WatchChan {
```
hunchback, line 93: prefix is opaque rather use:

```
func (con *DatabaseEtcd) AddWatchHandler(ctx context.Context, table, uuid string` or table string, uuid ...string)
```

roytman:
> @roytman a general comment is that I don't think the database abstraction is good for our use, I think every component which uses etcd should get a `*clientv3.Client` and then use clientv3 directly, I suggest this approach as our initial implementation will be tightly coupled to etcd, thus making the code simpler (as we don't attempt to provide a generic database interface).

@hunchback I remind you that you did this refactoring, without any internal discussions, now you again have decided by youself about  revert of database abstraction. This topic is unrelated to the review feature/PR

pkg/ovsdb/database.go
```diff
@@ -150,32 +159,64 @@ func (con *DatabaseEtcd) PutData(ctx context.Context, key string, obj interface{
...
+	fmt.Printf("DB Monitor 3 %v\n", ok)
+	if !ok {
+		// we don't have monitors for this database yet, create a new one.
+		m = newMonitor(dbName, con)
+		ctxt, cancel := context.WithCancel(context.Background())
+		m.cancel = cancel
+		wch := con.cli.Watch(clientv3.WithRequireLeader(ctxt),con.prefix + dbName,
+			clientv3.WithPrefix(),
+			clientv3.WithCreatedNotify(),
+			clientv3.WithPrevKV())
```
hunchback, line 178: this is a code duplication, as you have a specific way to create a monitor (for example with the PrevKV option set) better use a utility function, that way if in future you change options you will only have to modify one location

pkg/ovsdb/database.go
```diff
@@ -150,32 +159,64 @@ func (con *DatabaseEtcd) PutData(ctx context.Context, key string, obj interface{
...
+		m.watchChannel = wch
+		con.monitors[dbName] = m
+	}
+	fmt.Println("DB Monitor 4")
+	m.addUpdaters(con.prefix, updaters, handler)
+	fmt.Println("DB Monitor 5")
+	if !ok {
+		m.start()
 	}
+	fmt.Println("DB Monitor 6")
```
hunchback, line 188: maybe you want to remove some of the prints ...

pkg/ovsdb/database.go
roytman: Where do you see "the code duplication" ? only shared option is `clientv3.WithPrefix()` all others are watching related

pkg/ovsdb/database.go
roytman: yeap

pkg/ovsdb/database.go
```diff
@@ -150,32 +159,64 @@ func (con *DatabaseEtcd) PutData(ctx context.Context, key string, obj interface{
...
+func (con *DatabaseEtcd) AddMonitors(dbName string, updaters map[string][]*updater, handler handlerKey)  {
+	fmt.Println("DB Monitor 1")
 	con.mu.Lock()
 	defer con.mu.Unlock()
-	if monitor, ok := con.monitors[keysPrefix]; !ok {
-		con.monitors[keysPrefix] = *newMonitor(con.cli, con.prefix+keysPrefix, mcr, isV1, hand)
-	} else {
-		monitor.addHandler(mcr, isV1, hand)
+	fmt.Println("DB Monitor 2")
+	m, ok := con.monitors[dbName]
```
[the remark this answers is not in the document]
roytman: I have handlers (clients) counters

pkg/ovsdb/monitor_test.go
```diff
@@ -152,98 +130,55 @@ func TestTableUpdate(t *testing.T) {
...
-				mcr.Columns = col
-				mcr.Select = sel
-				cmp.MonitorCondRequests = map[string][]ovsjson.MonitorCondRequest{tableName: {mcr}}
-
-				monitorResponse, actualError := ch.Monitor(nil, cmp)
-				assert.Nilf(t, actualError, "handler call should not return error: %v", err)
-
-				monitorCondResponse, actualError := ch.MonitorCond(nil, cmp)
-				assert.Nilf(t, actualError, "handler call should not return error: %v", err)
+func TestMonitor(t *testing.T) {
```
hunchback, line 136: I rather that you break up this monolith test into a set of smaller tests each checking a specific feature, to eliminate code duplication just use utility functions. Basically you what that is a specific feature of monitor breaks then according to the test name you will be able to know what happened

pkg/ovsdb/monitor_test.go
```diff
@@ -152,98 +130,55 @@ func TestTableUpdate(t *testing.T) {
...
+	u11 := mcrToUpdater(mcr1, false)
+	m11 := map[string][]*updater{"table1":{u11}}
+	h11 := handlerKey{jasonValue: "jasonValue11"}
+	m.addUpdaters(dbPrefix, m11, h11)
+	assert.Equal(t, map[handlerKey]bool{h1:true, h2:true, h11:true}, m.handlers,"Handlers maps should be equals")
+	assert.Equal(t, map[string][]updater{"ovsdb/dbTest/table1":{*u1,*u2, *u11}, "ovsdb/dbTest/table2":{*u3}}, m.key2Updaters,"Key to updater maps should be equals")
+	assert.Equal(t, map[string][]handlerKey{u1.key:{h1, h2}, u2.key:{h1, h2}, u3.key:{h1, h2}, u11.key:{h11}}, m.upater2handlers)
+
+	m.removeUpdaters(dbPrefix, map[string][]string{"table1":{u11.key}}, h11)
+	assert.Equal(t, map[handlerKey]bool{h1:true, h2:true}, m.handlers,"Handlers maps should be equals")
```
hunchback, line 168: you are checking against the internal state in `m.handlers` this makes you tests verify sensitive to changes in implementation, and alternative would be to expose semantics which enable to get state externally which would enable you to implement the tests

pkg/ovsdb/database.go
roytman: done

pkg/ovsdb/database.go
hunchback: see `AddWatchHandler()`

pkg/ovsdb/database.go
hunchback:
```
gofmt -s -w *.go
```

pkg/ovsdb/handler.go
```diff
@@ -283,8 +258,12 @@ func (ch *Handler) SetConnection(con ClientConnection) {
 
 func (ch *Handler) notify(jsonValue interface{}, updates ovsjson.TableUpdates) {
 	var err error
-	nType := ch.jsonValueToUpdateNotificationType[jsonValue]
-	switch nType {
+	handler, ok := ch.monitors[jsonValue]
+	if !ok {
+		klog.Errorf("Unkown jasonValue %s", jsonValue)
```
hunchback, line 263: s/jason/json/

pkg/ovsdb/monitors.go
```diff
@@ -5,188 +5,239 @@ import (
...
 	Columns map[string]bool
 	Where   [][]string
+	Select  libovsdb.MonitorSelect
 	isV1    bool
+	key     string
 }
 
 type handlerKey struct {
 	handler    *Handler
 	jasonValue interface{}
```
hunchback, line 32: s/jasonValue/jsonValue/

pkg/ovsdb/monitors.go
```diff
@@ -5,188 +5,239 @@ import (
...
-				} else if ev.IsModify() {
-					dest = monitor.modify
-				} else {
-					// delete event
-					dest = monitor.delete
+		for wresp := range m.watchChannel {
+			if wresp.Canceled {
+				// TODO should we just reconnect
+				m.mu.Lock()
+				defer m.mu.Unlock()
```
hunchback, line 137: check lock

pkg/ovsdb/monitors.go
roytman: done

pkg/ovsdb/monitors.go
```diff
@@ -207,63 +271,78 @@ func (u *updater) prepareTableUpdate(event *clientv3.Event) (table string, uuid
...
+func unmarshalData(data []byte) (map[string]interface{}, error) {
+	obj := map[string]interface{}{}
+	if err := json.Unmarshal(data, &obj); err != nil {
+		return nil, err
+	}
+	delete(obj, "uuid")
+	return obj, nil
+}
+
+func (m *monitor) tableToKey(dbPrefix string, table string) string {
```
hunchback, line 333: dbPrefix should be renamed as its the prefix of the server/service

pkg/ovsdb/monitors.go
```diff
@@ -207,63 +271,78 @@ func (u *updater) prepareTableUpdate(event *clientv3.Event) (table string, uuid
...
+	return fmt.Sprintf("%s%s/%s", dbPrefix, m.dataBaseName, table)
+}
+
+func keyToTableAndUUID(key string) (path string, table string, uuid string, err error) {
```
hunchback, line 337: maybe define a type

```
type KeyParts struct {
        server string
        dbname string
        table string
        uuid string
}

File: pkg/ovsdb/monitors.go [hunk 13]
```diff
@@ -207,63 +271,78 @@ func (u *updater) prepareTableUpdate(event *clientv3.Event) (table string, uuid
...
+	return nil, nil
 }
 
-func keyToTableAndUUID(key string) (table string, uuid string, err error) {
+func unmarshalData(data []byte) (map[string]interface{}, error) {
+	obj := map[string]interface{}{}
+	if err := json.Unmarshal(data, &obj); err != nil {
+		return nil, err
+	}
+	delete(obj, "uuid")
```
hunchback, line 329: should it not be `_uuid` ? if so then maybe define in a constant

hunchback: also there are other specifcal fields that maybe you want to delete such as `_version`

pkg/ovsdb/database.go
roytman: AddWatchHandler is removed

pkg/ovsdb/monitor_test.go
roytman: I agree about the internal state, and sensitivity of the tests, but the internal state is quite complicated, and I want to be sure that it is updated correctly.

pkg/ovsdb/monitors.go
roytman: it is exactly dataBase prefix, but I plan to unify all these key names and their methods after merging of TX and Monitors. There is some duplication between these PRs

pkg/ovsdb/monitors.go
roytman: make sense, but again I'd  prefer to do during the keys unification

pkg/ovsdb/monitors.go
```diff
@@ -207,63 +271,78 @@ func (u *updater) prepareTableUpdate(event *clientv3.Event) (table string, uuid
...
+			// TODO compare sets and maps
+			if u.isV1 {
+				delete(prevData, column)
+			} else {
+				delete(data, column)
 			}
 		}
 	}
-	return
+	if !u.isV1 {
```
hunchback, line 312: can you refer me to non-v1 spec?

what I see currently is v1 which defines:

```
    "old": <row>   present for "delete" and "modify" updates
    "new": <row>   present for "initial", "insert", and "modify" updates

```

can we not start with v1 and then add additional version?

pkg/ovsdb/monitors.go
```diff
@@ -207,63 +271,78 @@ func (u *updater) prepareTableUpdate(event *clientv3.Event) (table string, uuid
...
 			}
 		}
-		if !u.isV1 {
-			if len(data) > 0 {
-				rowUpdate = &ovsjson.RowUpdate{Modify: &data}
-			}
-		} else {
-			if len(prevData) > 0 { // there are monitored updates
-				rowUpdate = &ovsjson.RowUpdate{New: &data, Old: &prevData}
+		if reflect.DeepEqual(cValue, prevData[column]) {
```
hunchback, line 303: you need to probably handle every single type comparison differently. and you may need the schema to know that type it actually is. (maybe add a XXX comment here).

pkg/ovsdb/monitors.go
roytman: Agree about uuid constant and about  _uuid. I've started to think, that maybe we don't need this (_uuid) column at all, we have is as part of the key.
`_version` is returned but OVSDB

pkg/ovsdb/monitors.go
roytman: https://docs.openvswitch.org/en/latest/ref/ovsdb-server.7/#update2-notification

pkg/ovsdb/monitors.go
```diff
@@ -5,198 +5,262 @@ import (
...
+				}
+				tableUpdate[uuid] = *rowUpdate
+			}
+		}
 	}
-	data := map[string]interface{}{}
+	return result, nil
+}
+
+func (u *updater) prepareRowUpdate(event *clientv3.Event) (*ovsjson.RowUpdate, error) {
```
hunchback, line 239: it is very difficult to verify that you have covered all cases as I don't see a clear structure here, you could for example due something along these lines (this is psaudo-code) using utility function to eliminate duplication.

```
switch event {
case "initial":
    switch select {
    case "insert":
        switch selectVersion {
        case "v1":
            switch columnType {
            case integer:
            case real:
            ...
            }
        case "v2":
        default:
            return errors.New("malformed select version")
        }
    case "delete"
    default:
        return errors.New("maformed select type")
    }
case "modify":
case "insert":
case "delete":
default:
    return errors.New("malformed event type")
}
```

pkg/ovsdb/monitors.go
```diff
@@ -5,198 +5,262 @@ import (
...
+				tableUpdate, ok := tableUpdates[table]
+				if !ok {
+					tableUpdate = ovsjson.TableUpdate{} // map[string]ovsjson.RowUpdate{}
+					tableUpdates[table] = tableUpdate
+				}
+				// check if there is a rowUpdate for the same uuid
+				_, ok = tableUpdate[uuid]
+				if ok {
+					// TODO fixme
+					klog.Warningf("Duplicate event for %s/%s/%s", m.dataBaseName, table, uuid)
```
hunchback, line 230: you should create the etcd key in a single place, so that when we do the adaptation you don't need to change several places.

pkg/ovsdb/monitors.go
```diff
@@ -5,198 +5,262 @@ import (
...
+
+func (m *monitor) prepareTableUpdate(events []*clientv3.Event) (map[handlerKey]ovsjson.TableUpdates, error) {
+	// prepare results that will be sent to clients
+	// handlerKey -> tableName -> uuid -> rowUpdate
+	result := map[handlerKey]ovsjson.TableUpdates{}
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	for _, ev := range events {
+		key := string(ev.Kv.Key)
+		path, table, uuid, err := keyToTableAndUUID(key)
```
hunchback, line 191: the path should be broken up into:
- server/service (e.g. ndb, sdb, etc)
- dbname (_Server, OVN_Southbound)

pkg/ovsdb/monitors.go
```diff
@@ -5,198 +5,262 @@ import (
...
+		if err != nil {
+			klog.Errorf("Wrong event's key %s", key)
```
hunchback, line 193: I am just wondering in what scenario can this actually happen?

roytman: it is just a guard for a wrong keys, for example, when the etcd keys generation was changed (add service name between ovsdb prefix and dbName).

pkg/ovsdb/monitors.go
roytman: it is not exactly the key, just warning message, but yes I agree

pkg/ovsdb/monitors.go
roytman: we cannot switch event, it has 2 methods isModified() and isCreated(), based on that we can separated different events, The version is boolean, so we don't need the default use case (and the switch)

pkg/ovsdb/monitors.go
```diff
@@ -5,198 +5,262 @@ import (
...
+		}
+		updaters, ok := m.key2Updaters[path]
+		if !ok {
+			klog.Infof("There is no monitors for table path %s", path)
+			continue
+		}
+		for _, updater := range updaters {
+			rowUpdate, err := updater.prepareRowUpdate(ev)
+			if err != nil {
+				klog.Errorf("prepareTableUpdate returned error %s, updater %v", err, updater)
```
hunchback, line 204: fix log message s/prepareTableUpdate/prepareRowUpdate/

pkg/ovsdb/monitors.go
```diff
@@ -207,63 +271,78 @@ func (u *updater) prepareTableUpdate(event *clientv3.Event) (table string, uuid
...
-	if l < 3 {
-		return "", "", fmt.Errorf("wrong formated key: %s", key)
+	if l < 4 {
+		err = fmt.Errorf("wrong formated key: %s", key)
+		return
 	}
-	return slices[l-2], slices[l-1], nil
+	path = fmt.Sprintf("%s/%s/%s", slices[0], slices[1], slices[2])
+	table = slices[l-2]
+	uuid = slices[l-1]
```
hunchback, line 346: can you detail what is the format you are handling here?
would be clearer if you would give names to each element:
```
x := slices[0]
y := slices[1]
z := slices[2]
```

pkg/ovsdb/monitors.go
```diff
@@ -5,198 +5,262 @@ import (
...
+func mcrToUpdater(mcr ovsjson.MonitorCondRequest, isV1 bool) *updater {
+	sort.Strings(mcr.Columns)
+	var key string
+	for _, c := range mcr.Columns {
+		key = key + c
+	}
+	// TODO handle "Where"
+	if mcr.Select == nil {
+		mcr.Select = &allTrueSelect
+	}
```
hunchback, line 178: I think it is per each option independantly, so whatever is missing get the value of true:

```
   o  If "initial" is omitted or true, every row in the table is sent as
      part of the response to the "monitor" request.

   o  If "insert" is omitted or true, "update" notifications are sent
      for rows newly inserted into the table.

   o  If "delete" is omitted or true, "update" notifications are sent
      for rows deleted from the table.

   o  If "modify" is omitted or true, "update" notifications are sent
      whenever a row in the table is modified.
```

PR #65, opened 2021-04-28
adding parameters to server and updating e2e

OmerBenHayun:


.travis.yml
```diff
@@ -37,7 +37,8 @@ jobs:
     - make install-tools
     before_script:
     - make etcd
-    - make server
+    - make north-server
+    - make south-server
```
roytman, line 41: I don't think for CI (travis) we need both NB and SB deployments. One of them should be enough

pkg/cmd/server/server.go
```diff
@@ -24,17 +26,25 @@ const UNIX_SOCKET = "/tmp/ovsdb-etcd.sock"
...
+	schemaFile         = flag.String("schema-file", "", "schema-file")
+	schemaName         = flag.String("schema-name", "", "schema-name")
+	loadServerDataFlag = flag.Bool("load-server-data", false, "load-server-data")
 )
 
 func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
+
+	fmt.Println("start the ovsdb-etcd server with the following arguments:")
```
roytman, line 44: It is good for development, and should be removed later. 
Please add the // TODO comment.

.travis.yml
OmerBenHayun: Thanks,Done

pkg/cmd/server/server.go
OmerBenHayun: Thanks,Done

PR #70, opened 2021-05-01
Unification of the etcd key processing 

roytman:


README.md
```diff
@@ -16,3 +16,16 @@ and you will find a description of the extensions beyond the spec here
...
+* all key parts are slash ("/") separated
+* keys start from deployment defined prefix, which can contain several slash separated sub-parts.
+* the next key part defines the deployment service, e.g. "sbdb" or "nbdb"
+* after that, we have a database server name, e.g. `OVN_Northbound`, `OVN_Southbound`, `_Server`. There  is an 
+  additional internal entry `_` for locks and comments.
+* next key element defines table names according to the relevant schema. Locks and comments are stored in the internal 
+  database in `_locks` and `_comments` entries.
+* the last key element is the `uuid` of the table row. This element can be the lock or comment id as well.
+* in order to guarantee that `_Server/Database` will contains only entries per running databases, its last key element
+  is not uuid, but the database server name.
```
hunchback, line 31: I would rather use semantics as follows to define (and below you can elaborate on each part):
- data:           <prefix>/<service>/<dbname>/<table>/<uuid> --> <row>
- locks:          <prefix>/<service>/_/_locks/<lockid> --> ?
- comments: <prefix>/<service>/_/_comments/<timestamp> --> <comment>

README.md
```diff
@@ -16,3 +16,16 @@ and you will find a description of the extensions beyond the spec here
...
+In the project we use the following etcd key formation convention:
+* all key parts are slash ("/") separated
+* keys start from deployment defined prefix, which can contain several slash separated sub-parts.
+* the next key part defines the deployment service, e.g. "sbdb" or "nbdb"
+* after that, we have a database server name, e.g. `OVN_Northbound`, `OVN_Southbound`, `_Server`. There  is an 
+  additional internal entry `_` for locks and comments.
+* next key element defines table names according to the relevant schema. Locks and comments are stored in the internal 
+  database in `_locks` and `_comments` entries.
+* the last key element is the `uuid` of the table row. This element can be the lock or comment id as well.
+* in order to guarantee that `_Server/Database` will contains only entries per running databases, its last key element
```
hunchback, line 30: I did not understand this comment. please explain.

pkg/cmd/server/server.go
```diff
@@ -43,12 +45,23 @@ func main() {
...
 	//TODO: for development purposes only. will be remove later
 	fmt.Println("start the ovsdb-etcd server with the following arguments:")
-	fmt.Printf("tcpAddress:%s\nunixAddressress:%s\netcdMembersress:%s\nschemaBasedir:%s\nmaxTasks:%d\ndatabasePrefix:%s\nschemaFile:%s\nschemaName:%s\nloadServerData:%v\n",
+	fmt.Printf("\ttcpAddress:%s\n\tunixAddressress:%s\n\tetcdMembersress:%s\n\tschemaBasedir:%s\n\tmaxTasks:%d\n\tdatabasePrefix:%s\n\tschemaFile:%s\n\tschemaName:%s\n\tloadServerData:%v\n",
 		*tcpAddress, *unixAddress, *etcdMembers, *schemaBasedir, *maxTasks, *databasePrefix, *schemaFile, *schemaName, *loadServerDataFlag)
 
 	if len(*tcpAddress) == 0 && len(*unixAddress) == 0 {
 		klog.Fatal("You must provide a network-address (TCP and/or UNIX) to listen on")
 	}
+	// several OVSDB ceployments can share the same etcd, but for rest of the work, we don't have to separate
```
hunchback, line 54: s/ceployments/deployments/

pkg/cmd/server/server.go
```diff
@@ -43,12 +45,23 @@ func main() {
...
+	// databasePrefix and serviceName.
+	if len(*databasePrefix) != 0 {
+		if len(*serviceName) != 0 {
```
hunchback, line 57: why are you introducing variation? why can't we just mandate a serviceName and make is standard?
databasePrefix should be renamed to ovsdbPrefix

pkg/cmd/server/server.go
```diff
@@ -43,12 +45,23 @@ func main() {
...
+			common.SetPrefix(*databasePrefix + common.KEY_DELIMETER + *serviceName)
```
hunchback, line 58: is this setting a global variable? if so that rather we don't

pkg/common/key.go
```diff
@@ -0,0 +1,110 @@
...
+	KEY_DELIMETER = "/"
+	LOCKS         = "_locks"
+	COMMENTS      = "_comments"
+	INTERNAL_DB   = "_"
+)
+
+var prefix string
+
+type Key struct {
+	Prefix    string
```
hunchback, line 20: Maybe the `Prefix` should be split:
- OvsdbPrefix (such as `ovsdb`)
- ServiceName (such as `ndb`)

pkg/common/key.go
```diff
@@ -0,0 +1,110 @@
...
+	DBName    string
+	TableName string
+	// the id represents uuid for the rows and id for the comments and locks
+	ID string
```
hunchback, line 24: can it UUID ... as it is also the `tablename` and `dbname` don't mean much when you store locks and comments ...

pkg/common/key.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func SetPrefix(prf string) {
+	prefix = prf
+}
+
+func GetPrefix() string {
+	return prefix
+}
```
hunchback, line 33: I rather you don't use a global variable ... it just makes multi-threading more difficult later

pkg/common/key.go
```diff
@@ -0,0 +1,110 @@
...
+		retKey = Key{Prefix: prefix, DBName: keyParts[l-3], TableName: keyParts[l-2], ID: keyParts[l-1]}
+	}
+	if retKey.DBName == "" || retKey.TableName == "" || retKey.ID == "" {
+		return nil, fmt.Errorf("wrong formatting key %s", keyStr)
+	}
+	if retKey.Prefix != prefix {
+		return nil, fmt.Errorf("wrong key, unmatched prefix %s, %s", retKey.Prefix, prefix)
+	}
+	return &retKey, nil
+}
```
hunchback, line 59: you tried to do a micro-optimization (making parts of the name optional) and thus resulting in code which is 10x more complicated. I think the cost out weights the benefits and we should rather stick to a standard representation of paths.

README.md
roytman: `_Server/Database` contains a row per running database, for example for NB server it will contain 2 entries one for OVN_Northbound and another for _Server.
Part of the entries (rows) are updated per each server start. If we use a standard key schema, uuid as a row kew, we have to clean the entries each time when a server is starting. When we run several instances of our server, prevention of creation multiple entires will create huge headache.

pkg/cmd/server/server.go
roytman: yeap

pkg/cmd/server/server.go
roytman: ovsdbPrefix - OK,
databaseName  - is mandatory
I prefer do not request serviceName  and ovsdbPrefix as mandatory.

pkg/cmd/server/server.go
roytman: Yes it is a global variable and it is set during the server startup (one of the common patterns like logging), otherwise, we have to have a pointer to some commonplace with this configuration in each place.

pkg/common/key.go
roytman: why do we need to split it here? Once the variables are set the server does not need to separate them.

pkg/common/key.go
roytman: `DBName` does mean much for locks and comments. They are per database. 
TableName -? but you can as what are the table in key/value store

pkg/common/key.go
roytman: 1. We do support multi-threading now. Multiple clients and even the same client can send simultaneous requests to the server. 
 2. This variable is set during server initialization, before opening any sockets.
 3. It is one of the standard server design patterns 
 4. I would prefer do not use global variables either, but it would significantly complicate the design, and implementation . Where is your washing to simplicity?

pkg/common/key.go
roytman: It is not a micro-optimization, but supporting different user configuration options. You are talking about 10x more complicated code, but it is a single not too complicated function. 
In your scenario you would require from users to provide ovsdbPrefix, serviceName - and btw none of them cannot contain slashes. Think about explanation about all these, relative to the function above.

pkg/common/key.go
hunchback: and you can add then helper functions:

```
func NewKey(prefix, service, dbname, table string, uuid ...string) {
}
```

```
func NewCommentKey(prefix string) {
}
```

```
func NewLockKey(string prefix, service, ...) {
}
````

roytman: the variable arguments list creates possibilities for an errors, e.g. we will have to check if a function received more than a single var argument, and we will have to return an error, and to check it in all places where we call the functions...
For each type (data, comment and lock) I provide 2 functions with the same internal implementation.

- `func NewCommentKey(commentID string)`
-  `func NewCommentTableKey()`
and btw, if a user calls the first function with commentID == "", it will equal to the second (table key ) function.
Added comments about the functions

PR #71, opened 2021-05-02
Txn schema

hunchback:


pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+atomic fields).
+*/
+
+//ExtendedType includes atomic types as defined in the RFC plus Enum, Map and Set
+type ExtendedType = string
+
+// RefType is used to define the possible RefTypes
+type RefType = string
+
+const (
```
roytman, line 102: constants should be uppercase

pkg/libovsdb/schema.go
```diff
@@ -1,29 +1,37 @@
 package libovsdb
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 )
 
+type Schema map[string]*DatabaseSchema
```
roytman, line 10: why do you need this `Shema` type. if you want to store several schemas in the server you can define it there. 
RFC7047 Schema definition is exactly `DataBaseSchema`

pkg/libovsdb/schema.go
```diff
@@ -1,29 +1,37 @@
...
+
 // DatabaseSchema is a database schema according to RFC7047
 type DatabaseSchema struct {
 	Name    string                 `json:"name"`
 	Version string                 `json:"version"`
 	Tables  map[string]TableSchema `json:"tables"`
```
roytman, line 16: Checksum is missing `json:"cksum,omniempty"`

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
 	}
 	return true
 }
+
+// TableSchema is a table schema according to RFC7047
+type TableSchema struct {
+	Columns map[string]*ColumnSchema `json:"columns"`
+	Indexes [][]string               `json:"indexes,omitempty"`
```
roytman, line 86: missing
"maxRows": <integer>                      optional
"isRoot": <boolean>                       optional

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
 	}
 	return true
 }
+
+// TableSchema is a table schema according to RFC7047
+type TableSchema struct {
+	Columns map[string]*ColumnSchema `json:"columns"`
```
roytman, line 85: why do you use pointer to ColumnSchema ?

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+	return strings.Join([]string{typeStr, flagStr}, " ")
+}
+
+// UnmarshalJSON unmarshalls a json-formatted column
+func (column *ColumnSchema) UnmarshalJSON(data []byte) error {
+	// ColumnJSON represents the known json values for a Column
+	type ColumnJSON struct {
+		TypeRawMsg json.RawMessage `json:"type"`
+		Ephemeral  bool            `json:"ephemeral,omitempty"`
+		Mutable    bool            `json:"mutable,omitempty"`
```
roytman, line 223: if Mutable is not specified, it will be `false`, but I think it will incorrectly translate the RFC.

> If "mutable" **is specified** as false, then this column's values may
      not be modified after they are initially set with the "insert"
      operation.
In the current schemas, I don't see this flag at all, so the default should be `true`

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+			return err
+		}
+	}
+
+	if !isAtomicType(column.TypeObj.Key.Type) {
+		return fmt.Errorf("schema contains unknown atomic type %s", column.TypeObj.Key.Type)
+	}
+
+	// 'value' is optional. If it exists, we know the real native type is a map
+	if colTypeJSON.ValueRawMsg != nil {
```
roytman, line 314: if it can simplify, the map's key are strings

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+// BaseType is a base-type structure as per RFC7047
+type BaseType struct {
+	Type string `json:"type"`
+	// Enum will be parsed manually and set to a slice
+	// of possible values. They must be type-asserted to the
+	// corret type depending on the Type field
+	Enum       []interface{} `json:"_"`
+	MinReal    float64       `json:"minReal,omitempty"`
+	MaxReal    float64       `json:"maxReal,omitempty"`
+	MinInteger int           `json:"minInteger,omitempty"`
```
roytman, line 165: I would add comments the comments from the RFC to the BaseType constrains, e.g.

> "minInteger": <integer>          optional, **integers only**
         "maxInteger": <integer>          optional, **integers only**
         "minReal": <real>                optional, **reals only**
         "maxReal": <real>                optional, **reals only**
         "minLength": <integer>           optional, **strings only**
         "maxLength": <integer>           optional, **strings only**
         "refTable": <id>                 optional, **UUIDs only**
         "refType": "strong" or "weak"    optional, o**nly with "refTable"**
 this can be especially useful for strings and refs

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+	MaxInteger int           `json:"maxInteger,omitempty"`
+	MinLength  int           `json:"minLength,omitempty"`
+	MaxLength  int           `json:"maxLength,omitempty"`
+	RefTable   string        `json:"refTable,omitempty"`
+	RefType    RefType       `json:"refType,omitempty"`
+}
+
+// String returns a string representation of the (native) column type
```
roytman, line 173: you create some kind of type encoding here, a comment with explanation can be useful

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+	case TypeString:
+		return ""
+	case TypeUUID:
+		return UUID{"00000000-0000-0000-0000-000000000000"}
+	default:
+		panic(fmt.Sprintf("Unsupported type %s", columnSchema.Type))
+	}
+}
+
+func (tableSchema TableSchema) Default(row *map[string]interface{}) {
```
roytman, line 400: should it be  (tableSchema *TableSchema)

pkg/libovsdb/schema.go
```diff
@@ -71,3 +79,344 @@ func (schema DatabaseSchema) validateOperations(operations ...Operation) bool {
...
+
+func (databaseSchema DatabaseSchema) Default(table string, row *map[string]interface{}) {
+	tableSchema, ok := databaseSchema.Tables[table]
+	if !ok {
+		panic(fmt.Sprintf("Missing schema for table %s", table))
+	}
+	tableSchema.Default(row)
+}
+
+func (schema Schema) Default(dbname, table string, row *map[string]interface{}) {
```
roytman, line 416: it is not part of the schema processing

pkg/libovsdb/schema_test.go
```diff
@@ -0,0 +1,352 @@
+package libovsdb
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
```
roytman, line 7: you don't need this empty line

pkg/libovsdb/schema.go
hunchback: ok

pkg/libovsdb/schema.go
hunchback: ok

pkg/libovsdb/schema.go
hunchback: this is the original code from libovsdb

pkg/libovsdb/schema.go
hunchback: renamed to `Schemas`

pkg/libovsdb/schema.go
hunchback: original code (I generally agree with your comment)

pkg/libovsdb/schema.go
hunchback: agree

pkg/libovsdb/schema.go
hunchback: libovsdb code

pkg/libovsdb/schema.go
hunchback: added to strings and refs (others have the type in there name)

pkg/libovsdb/schema.go
hunchback: ok

pkg/libovsdb/schema.go
hunchback: as explained above, its an integral part of managing schemas for multiple databases. Renamed to `Schemas` as a greed

pkg/libovsdb/schema_test.go
hunchback: ok (was original code)

PR #72, opened 2021-05-04
txn: added schema to conditions

hunchback:


pkg/ovsdb/transact.go
```diff
@@ -377,14 +378,20 @@ func NewCondition(condition []interface{}) (*Condition, error) {
 	if !ok {
 		return nil, errors.New(E_INTERNAL_ERROR)
 	}
+	schema := tableSchema.LookupColumn(column)
```
roytman, line 381: `schema` name can be misleading, what about `columnSchema` or `cSchema` ...

roytman:
From previous PR, please add `omitempt` to "Checksm string                 `json:"checksm"`"

pkg/ovsdb/transact.go
```diff
@@ -366,9 +366,10 @@ type Condition struct {
 	Column   string
 	Function string
 	Value    interface{}
+	Schema   *libovsdb.ColumnSchema
```
roytman, line 369: s/Schema/ColumnSchema/

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

PR #89, opened 2021-05-07
Fix transaction operation response objects

roytman:


hunchback (changes requested):
do not merge - see alternative implementation using pointers rather than multiple structs

roytman:
At the first glance, I see at least one problem. When, you run select the operation and nothing to return, you have to return empty Rows set,  in the case of the others operations, this empty set creates parsing errors.

PR #90, opened 2021-05-08
libovsdb: OperationResult fields optional

hunchback:


pkg/ovsdb/transact.go
```diff
@@ -228,7 +228,10 @@ func NewTransaction(cli *clientv3.Client, request *libovsdb.Transact) *Transacti
 	txn.cache = Cache{}
 	txn.schemas = libovsdb.Schemas{}
 	txn.request = *request
-	txn.response.Result = make([]libovsdb.OperationResult, len(request.Operations))
+	txn.response.Result = make([]*libovsdb.OperationResult, len(request.Operations))
+	for i, _ := range txn.response.Result {
+		txn.response.Result[i] = libovsdb.NewOperationResult()
```
roytman, line 233: This `NewOperationResult` call creates an instance of OperationResult which marshals  to JSON string as 
{"count":0,"error":"","details":"","uuid":["named-uuid",""],"rows":null}.
In order to be inline with the spec, I suggest just to create an empty OperationResult {}, and during the operations, just initialize the relevant field.

pkg/libovsdb/notation.go
```diff
@@ -149,17 +149,27 @@ func NewTransact(params []interface{}) (*Transact, error) {
...
 // OperationResult is the result of an Operation
 type OperationResult struct {
-	Count   int         `json:"count,omitempty"`
-	Error   string      `json:"error,omitempty"`
-	Details string      `json:"details,omitempty"`
-	UUID    UUID        `json:"uuid,omitempty"`
-	Rows    []ResultRow `json:"rows,omitempty"`
+	Count   *int         `json:"count,omitempty"`
+	Error   *string      `json:"error,omitempty"`
+	Details *string      `json:"details,omitempty"`
```
roytman, line 160: There is no `Details` filed

pkg/ovsdb/transact.go
```diff
@@ -1020,8 +1023,8 @@ func doSelect(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.O
 		if err != nil {
 			return err
 		}
-		ovsResult.Count = ovsResult.Count + 1
-		ovsResult.Rows = append(ovsResult.Rows, *resultRow)
+		*ovsResult.Count++
```
roytman, line 1026: before going into the for loop, we have to create the rows set.
*ovsResult.Rows = []libovsdb.Row{}
you don't need Count here

pkg/ovsdb/transact.go
```diff
@@ -1045,11 +1048,11 @@ func doUpdate(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.O
...
-			ovsResult.Error = err.Error()
+			*ovsResult.Error = err.Error()
 			return err
 		}
 
-		ovsResult.Count = ovsResult.Count + 1
+		*ovsResult.Count++
 		err = RowUpdate(tableSchema, row, ovsOp.Row)
 		if err != nil {
 			return err
```
roytman, line 1058: you just return errors, in each operation, but you have to set the Erro filed of the specific operation

pkg/ovsdb/transact.go
```diff
@@ -1114,7 +1117,7 @@ func doDelete(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.O
 func preWait(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.OperationResult) error {
 	if ovsOp.Timeout != 0 {
 		err := fmt.Errorf("only support timeout 0")
-		ovsResult.Error = err.Error()
+		*ovsResult.Error = err.Error()
```
roytman, line 1120: this error set should be done for all operations

roytman (changes requested):

pkg/libovsdb/client.go
```diff
@@ -299,8 +299,8 @@ func (ovs OvsdbClient) MonitorCancel(jsonContext interface{}) error {
 	if err != nil {
 		return err
 	}
-	if reply.Error != "" {
-		return fmt.Errorf("Error while executing transaction: %s", reply.Error)
+	if *reply.Error != "" {
```
roytman, line 302: It's a part of the libovsdb, but we don't have to use OperationResult here, just return an error

pkg/ovsdb/transact.go
```diff
@@ -980,7 +983,7 @@ func doInsert(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.O
 		for _, row := range txn.cache.Table(txn.request.DBName, ovsOp.Table) {
 			if (*row)[COL_UUIDNAME] == ovsOp.UUIDName {
```
roytman, line 984: UUIDName can't be part of the cache, and the key, it is not UUID, and is provided only to  connect to different entries in the request .

Your logic here is wrong, you first, bring entire table, but when did not check the new UUID with the  cache

pkg/ovsdb/transact.go
```diff
@@ -989,11 +992,11 @@ func doInsert(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.O
 	ok := txn.schemas.Validate(txn.request.DBName, ovsOp.Table, &ovsOp.Row)
 	if !ok {
 		err := errors.New(E_CONSTRAINT_VIOLATION)
-		ovsResult.Error = err.Error()
+		*ovsResult.Error = err.Error()
 		return err
 	}
 
-	ovsResult.Count = ovsResult.Count + 1
+	*ovsResult.Count++
```
roytman, line 999: Insert doesn't return Count, only UUID or error

pkg/ovsdb/transact.go
```diff
@@ -1020,8 +1023,8 @@ func doSelect(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.O
 		if err != nil {
```
roytman, line 1023: 1. we have to set error in the operation result,
2.  we have to clean Row there

pkg/ovsdb/transact.go
hunchback: ok

pkg/libovsdb/notation.go
hunchback: it was originally there and it is used in libovsdb/examples

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

pkg/libovsdb/client.go
hunchback: will leave as is

pkg/ovsdb/transact.go
hunchback: ok

pkg/ovsdb/transact.go
hunchback: ok

PR #91, opened 2021-05-08
Txn schema enum

hunchback:


pkg/libovsdb/schema.go
```diff
@@ -541,34 +545,110 @@ func (schemas *Schemas) LookupTable(dbname, table string) *TableSchema {
...
-	_, ok := value.(bool)
-	return ok
+	typeval, ok := value.(bool)
+	if !ok {
+		return false
+	}
+	if baseType.Enum == nil {
+		return true
+	}
+	for _, v := range baseType.Enum {
```
roytman, line 595: I don't think that boolean can contain Enum, but it OK to have it just in case

pkg/ovsdb/transact.go
```diff
@@ -565,6 +582,8 @@ func (c *Condition) Compare(row *map[string]interface{}) (bool, error) {
 		return c.CompareString(row)
 	case libovsdb.TypeUUID:
 		return c.CompareUUID(row)
+	case libovsdb.TypeEnum:
```
roytman, line 585: Why do we need to compare Enum with Condition? I think Enum is relevant as a restriction to new values only, isn't it ?

pkg/libovsdb/schema.go
hunchback: removed

pkg/ovsdb/transact.go
hunchback: if enum then you need to check what type it is an compare accordingly (we support only enum strings for now)

PR #96, opened 2021-05-10
transact: fix compare for _uuid and _version

hunchback:


pkg/ovsdb/transact.go
```diff
@@ -551,6 +555,13 @@ func (c *Condition) CompareMap(row *map[string]interface{}) (bool, error) {
 }
 
 func (c *Condition) Compare(row *map[string]interface{}) (bool, error) {
+	switch c.Column {
+	case COL_UUID:
+		return c.CompareUUID(row)
+	case COL_VERSION:
```
roytman, line 561: Version type is UUID 

>  "_version": Like "_uuid", this column contains exactly one UUID
         value, initialized to a random value by the database engine
         when it creates a row, and it is read-only.  However, its value
         changes to a new random value whenever any other field in the
         row changes.  Furthermore, its value is ephemeral: when the
         database is closed and reopened, or when the database process
         is stopped and then started again, each "_version" also changes
         to a new random value.

PR #106, opened 2021-05-13
Clean up the OVNk deployment files

roytman:
- Remove unused files
- Add environment variables 
- Add installation of etcd and ovs/ovn packages into the images, instead of file copies. 


Makefile
```diff
@@ -1,8 +1,9 @@
 ROOT_DIR := .
-SERVER_EXECUTEBLE=$(ROOT_DIR)/pkg/cmd/server/server
+SERVER_EXECUTEBLE := "$(ROOT_DIR)/pkg/cmd/server/server"
 
-OVN_KUBERNETES_ROOT := \
-	~/work/src/github.com/ovn-org/ovn-kubernetes
+ifndef OVN_KUBERNETES_ROOT
+override OVN_KUBERNETES_ROOT := "$(ROOT_DIR)/../../ovn-org/ovn-kubernetes"
+endif
```
hunchback, line 6: use 
```
OVN_KUBERNETERS_ROOT ?= $(ROOT_DIR)/../../ovn-org/ovn-kubernetes
```

Makefile
```diff
@@ -51,6 +52,10 @@ verify: $(VERIFY)
 etcd:
 	$(MAKE) -C tests/e2e/ etcd &
 
+.PHONY: etcd-travis
+etcd-travis:
+	$(MAKE) -C tests/e2e/ etcd &
```
hunchback, line 57: why? this is code duplication

roytman: as we discussed today, I want to separate travis directives from other make directives

Makefile
```diff
@@ -67,6 +72,11 @@ north-server:
 south-server:
 	$(MAKE) -C tests/e2e/ server -e TCP_ADDRESS=:6642 UNIX_ADDRESS=/tmp/ovnsb_db.db DATABASE-PREFIX=ovsdb SERVICE-NAME=sb SCHEMA-FILE=ovn-sb.ovsschema LOAD-SERVER-DATA=FALSE PID-FILE=/tmp/sb-ovsdb.pid &
 
+.PHONY: north-server-travis
+north-server-travis:
+	$(MAKE) -C tests/e2e/ server -e TCP_ADDRESS=:6641 UNIX_ADDRESS=/tmp/ovnnb_db.db DATABASE-PREFIX=ovsdb SERVICE-NAME=nb SCHEMA-FILE=ovn-nb.ovsschema LOAD-SERVER-DATA=FALSE PID-FILE=/tmp/nb-ovsdb.pid &
```
hunchback, line 77: use Makefile macros:
```
# ndb can be run with : $(call run_server,nb)
# sdb can be run with : $(call run_server,sb)
define run_server
	$(MAKE) -C tests/e2e/ server -e TCP_ADDRESS=:6641 UNIX_ADDRESS=/tmp/ovnnb_db.db DATABASE-PREFIX=ovsdb SERVICE-NAME=$1 SCHEMA-FILE=ovn-$1.ovsschema LOAD-SERVER-DATA=FALSE PID-FILE=/tmp/nb-ovsdb.pid 
endef
```

Makefile
roytman: it doesn't work. I checked it

roytman: I want to allow to set OVN_KUBERNETERS_ROOT as an environment variable, plus there is a default value

hunchback:
```
vagrant@vagrant:~/projects/openvswitch$ unset OVN_KUBERNETES_ROOT
vagrant@vagrant:~/projects/openvswitch$ make
echo xxx/../../ovn-org/ovn-kubernetes
xxx/../../ovn-org/ovn-kubernetes
vagrant@vagrant:~/projects/openvswitch$ cat Makefile
ROOT_DIR=xxx
OVN_KUBERNETES_ROOT ?= $(ROOT_DIR)/../../ovn-org/ovn-kubernetes

all:
        echo ${OVN_KUBERNETES_ROOT}

vagrant@vagrant:~/projects/openvswitch$
```

Makefile
hunchback: ok, then I misunderstood you ... and now that I understand that you are simply duplicate all entries then I object to this ... may understanding was that in cases that the travis target differs from the standard target then you create a travis specific, the example you gave me was:
- travis - running in background
- during development - running in foreground

the duplication the never beneficial as it means that you can never stabilize the project (you fix one place but the other is still broken).

so now that I understand what you are doing, please explain what what is the benefit of duplication which out weighs the obvious downside.

Makefile
roytman: ok, and now try to do the same when you set OVN_KUBERNETES_ROOT

roytman: ok

Makefile
roytman: Will be a separate PR

Makefile
roytman: Will be a separate PR

PR #141, opened 2021-05-24
txn: split transactions with above 128 operations

hunchback:


roytman:
This breaks the transaction concept, I don't think we need this patch.

pkg/ovsdb/transact.go
```diff
@@ -411,6 +417,55 @@ func (mapUUID MapUUID) ResolvRow(row *map[string]interface{}) error {
...
+	res, err := etcd.Cli.Txn(etcd.Ctx).If(etcd.If...).Then(etcd.Then...).Else(etcd.Else...).Commit()
+	if err != nil {
+		return err
+	}
+	etcd.Res = res
+	return nil
+}
+
+func (etcd *Etcd) Split() []*Etcd {
+	ETCD_MAX_TXN_OPS := 128
```
roytman, line 455: Pls, put it as a separate constant, if we work with splitting tnx, we probably will want to configure it.

PR #174, opened 2021-06-01
wait: fix implementation and add tests

hunchback:


pkg/ovsdb/transact.go
```diff
@@ -1556,52 +1513,77 @@ func preWait(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.Op
...
+	if ovsOp.Columns == nil {
+		klog.Errorf("missing columns parameter")
+		return errors.New(E_CONSTRAINT_VIOLATION)
+	}
+
+	if ovsOp.Until == nil {
+		klog.Errorf("missing until parameter")
+		return errors.New(E_CONSTRAINT_VIOLATION)
+	}
+
```
roytman, line 1537: should you check that ovsOp.Where is not nil ? ( it is a required parameter too)

pkg/ovsdb/transact.go
```diff
@@ -1556,52 +1513,77 @@ func preWait(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.Op
...
+				return err
+			}
+			if equal && cond {
+				return nil
+			}
+		}
 	}
 
-	if ovsOp.Rows == nil {
+	if !equal {
```
roytman, line 1581: if equal is "false" but the cond  is "true", you return nil, but it should be "time-out" error

pkg/ovsdb/transact.go
```diff
@@ -1556,52 +1513,77 @@ func preWait(txn *Transaction, ovsOp *libovsdb.Operation, ovsResult *libovsdb.Op
...
-	case FN_NE:
-		if equal {
-			return errors.New(E_TIMEOUT)
-		}
-	default:
-		return errors.New(E_CONSTRAINT_VIOLATION)
-	}
-	return nil
+	klog.Errorf("wait: timed out")
+	return errors.New(E_TIMEOUT)
```
roytman, line 1586: can you validate, that if the wait operation returns "timeout error" the other transaction operation are not executed ?

pkg/ovsdb/transact_test.go
```diff
@@ -1042,24 +1042,143 @@ func TestTransactDelete(t *testing.T) {
...
+			},
+		},
+	}
+	common.SetPrefix("ovsdb/nb")
+	testEtcdCleanup(t)
+	resp, _ := testTransact(t, req)
+	assert.Nil(t, resp.Error)
+}
+
+func TestTransactWaitError(t *testing.T) {
```
roytman, line 1143: related to my comment about equal is "false", you don't have a test where until is FN_NE, but the rows are equals.

pkg/ovsdb/transact.go
hunchback: it will fail this specific operation and as a result will fail the transaction before reaching the 2nd Commit.

pkg/ovsdb/transact.go
hunchback: good catch, fixed

pkg/ovsdb/transact_test.go
hunchback: agree, added

pkg/ovsdb/transact.go
hunchback: I assume if it does not exist that means everything

PR #176, opened 2021-06-02
[WIP] Remove jsonValueStr, use array instead of map

roytman:


roytman:
We have a different implementation

PR #209, opened 2021-06-09
txn: fix klogr output

hunchback:


OmerBenHayun:
please use `make verify` and amend the results (so the CI build test will pass).
as for the content - It looks good to me.

PR #213, opened 2021-06-10
txn: add address and id from jsonrpc peer

hunchback:


roytman:
I'd suggest to unify all handler calls to print id

PR #223, opened 2021-06-14
server: use klogr logging

hunchback:


roytman:
The server fails with the following message 
flag provided but not defined: -logtostderr

PR #247, opened 2021-06-21
make: improved ovnkube targets

hunchback:


roytman:
in order to fetch log files, I prefer to use a script like 
[getLogs.zip](https://github.com/IBM/ovsdb-etcd/files/6691470/getLogs.zip)

PR #248, opened 2021-06-21
Add map/set processing for update rows

roytman:

### DIFF
--- a/pkg/ovsdb/handler.go
+++ b/pkg/ovsdb/handler.go
@@ -13,7 +13,7 @@ import (
 	"github.com/ibm/ovsdb-etcd/pkg/common"
 	"github.com/ibm/ovsdb-etcd/pkg/libovsdb"
 	"github.com/ibm/ovsdb-etcd/pkg/ovsjson"
-	shortuuid "github.com/lithammer/shortuuid/v3"
+	"github.com/lithammer/shortuuid/v3"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/concurrency"
 	"k8s.io/klog/v2"
@@ -196,7 +196,7 @@ func (ch *Handler) Steal(ctx context.Context, param interface{}) (interface{}, e
 }
 
 func (ch *Handler) MonitorCond(ctx context.Context, params []interface{}) (interface{}, error) {
-	ch.log.V(1).Info("monitorCond request", "params", params)
+	ch.log.V(5).Info("monitorCond request", "params", params)
 	updatersMap, err := ch.addMonitor(params, ovsjson.Update2)
 	if err != nil {
 		ch.log.Error(err, "monitorCond from remote")
@@ -255,14 +255,22 @@ func (ch *Handler) MonitorCondChange(ctx context.Context, params []interface{}) 
 		if !ok {
 			ch.log.Info("MonitorCondChange there is no monitor", "dbname", monitorData.dataBaseName)
 		}
+		databaseSchema, ok := ch.db.GetSchemas()[dbName]
+		if !ok {
+			return nil, fmt.Errorf("there is no databaseSchema for %s", dbName)
+		}
 		for tableName, mcrArray := range mcrs {
 			key := common.NewTableKey(dbName, tableName)
 			_, ok := monitor.key2Updaters[key]
 			if !ok {
 				ch.log.V(6).Info("MonitorCondChange", "table", tableName, "mcr", mcrArray)
 				var updaters []updater
+				tableSchema, err := databaseSchema.LookupTable(tableName)
+				if err != nil {
+					return nil, err
+				}
 				for _, mcr := range mcrArray {
-					updater := mcrToUpdater(mcr, jsonValueString, monitorData.notificationType == ovsjson.Update)
+					updater := mcrToUpdater(mcr, jsonValueString, tableSchema, monitorData.notificationType == ovsjson.Update)
 					updaters = append(updaters, *updater)
 				}
 				monitorData.updatersKeys = append(monitorData.updatersKeys, key)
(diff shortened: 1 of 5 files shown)


pkg/ovsdb/handler.go
```diff
@@ -255,14 +255,22 @@ func (ch *Handler) MonitorCondChange(ctx context.Context, params []interface{})
 		if !ok {
 			ch.log.Info("MonitorCondChange there is no monitor", "dbname", monitorData.dataBaseName)
 		}
+		schema, ok := ch.db.GetSchemas()[dbName]
```
hunchback, line 258: rename `schema` to `databaseSchema` so it is clear which one

pkg/ovsdb/handler.go
```diff
@@ -255,14 +255,22 @@ func (ch *Handler) MonitorCondChange(ctx context.Context, params []interface{})
...
+		if !ok {
+			return nil, fmt.Errorf("there is no schema for %s", dbName)
+		}
 		for tableName, mcrArray := range mcrs {
 			key := common.NewTableKey(dbName, tableName)
 			_, ok := monitor.key2Updaters[key]
 			if !ok {
 				ch.log.V(6).Info("MonitorCondChange", "table", tableName, "mcr", mcrArray)
 				var updaters []updater
+				tableSchema, ok := schema.Tables[tableName]
```
hunchback, line 268: You can use the helper function LookupTable() rather than map lookup, it provides am error string

pkg/ovsdb/monitor.go
```diff
@@ -251,13 +252,14 @@ func (m *dbMonitor) cancelDbMonitor() {
...
 
-func mcrToUpdater(mcr ovsjson.MonitorCondRequest, jsonValue string, isV1 bool) *updater {
+func mcrToUpdater(mcr ovsjson.MonitorCondRequest, jsonValue string, tableSchema *libovsdb.TableSchema, isV1 bool) *updater {
 
 	//TODO  handle "Where"
 	if mcr.Select == nil {
 		mcr.Select = &libovsdb.MonitorSelect{}
 	}
-	return &updater{Columns: common.StringArrayToMap(mcr.Columns), jasonValueStr: jsonValue, isV1: isV1, Select: *mcr.Select}
+	columns := columnArrayToMap(mcr.Columns, tableSchema.Columns)
```
hunchback, line 261: why don't you transfer `tableSchema` itself?

pkg/ovsdb/monitor.go
```diff
@@ -480,3 +527,45 @@ func (u *updater) prepareRow(value []byte) (map[string]interface{}, string, erro
...
+func difference(set1 libovsdb.OvsSet, set2 libovsdb.OvsSet) libovsdb.OvsSet {
+	var diff libovsdb.OvsSet
+
+	// Loop two times, first to find slice1 strings not in slice2,
+	// second loop to find slice2 strings not in slice1
+	for i := 0; i < 2; i++ {
+		for _, s1 := range set1.GoSet {
+			found := false
+			for _, s2 := range set2.GoSet {
+				if s1 == s2 {
```
msherif1234, line 554: is there relationship between s1 and s2 i.e one of them is subset of the other ? because here u check for the 1st matching element and stop there could be another mismatching fields after ?

roytman: the sets do not have any relationship, and I check them 2, see the swap operation at the end of the function. The break doesn't stop on the first matching element, it just breaks the internal loop.
See the tests that I just added to the monitor_test.go.
 
P.S. the PR is not ready

pkg/ovsdb/handler.go
```diff
@@ -409,12 +418,20 @@ func (ch *Handler) addMonitor(params []interface{}, notificationType ovsjson.Upd
 	if _, ok := ch.handlerMonitorData[jsonValueString]; ok {
 		return nil, fmt.Errorf("duplicate monitor ID")
 	}
+	schema, ok := ch.db.GetSchemas()[cmpr.DatabaseName]
```
hunchback, line 421: please rename `schema` to `databaseSchema`

pkg/ovsdb/handler.go
```diff
@@ -409,12 +418,20 @@ func (ch *Handler) addMonitor(params []interface{}, notificationType ovsjson.Upd
...
+	if !ok {
+		return nil, fmt.Errorf("there is no schema for %s", cmpr.DatabaseName)
+	}
 	updatersMap := Key2Updaters{}
 	var updatersKeys []common.Key
 	for tableName, mcrs := range cmpr.MonitorCondRequests {
 		var updaters []updater
+		tableSchema, ok := schema.Tables[tableName]
```
hunchback, line 429: please use `LookupTable()`

pkg/ovsdb/handler.go
hunchback: You may consider adding to libovsdb a method `LookupDatabase()` which already returns a correctly formatted error in case of failure

pkg/ovsdb/monitor.go
```diff
@@ -24,7 +25,7 @@ const (
 )
 
 type updater struct {
-	Columns          map[string]bool
+	Columns          map[string]*libovsdb.ColumnSchema
```
hunchback, line 28: Basically if you hold the TablesSchema you get the exact functionality you need by using `LookupColumn()` w/o the need to create/maintain the Columns field

pkg/ovsdb/handler.go
roytman: in general, I think we need to change the GetSchemas() method. In transactions, you don't need all schemas too, transactions always work with a specific database.

pkg/ovsdb/monitor.go
roytman: I changed the monitor internals, and removed the map at all.

PR #253, opened 2021-06-23
Txn events print

hunchback:


roytman:
What do you think to extend the events print to beyond transactions, so monitor notification will be abelto use them too?

Issue #259, opened 2021-06-24
Wrong delete  mutation


roytman:
Transaction:

> {"dbname":"OVN_Northbound","operations":[{"op":"mutate","table":"Logical_Router",     "mutations":[["ports","delete",["set",[["uuid","0535086f-f347-4bee-a130-fb24034174e0"]]]],["ports","insert",["set",[["named-uuid","row6dc0b1cf_7554_4a41_a57e_2f6d51852632"]]]]],"where":[["_uuid","==",["uuid","d2b112f2-56b2-41bf-b9ce     -4c1641b93bf4"]]]},{"op":"insert","table":"Logical_Router_Port","row":{"mac":"0a:58:64:40:00:01","name":"rtoj-ovn_cluster_router","networks":"100.64.0.1/16"},"uuid-name":"row6dc0b1cf_7554_4a41_a57e_2f6d51852632"},{"op":"comment","co     mment":"ovs-nbctl: run -- --if-exists lrp-del rtoj-ovn_cluster_router -- lrp-add ovn_cluster_router rtoj-ovn_cluster_router 0a:58:64:40:00:01 100.64.0.1/16"}]}

there is no delete etcd key message

notification events:

[...]

PR #269, opened 2021-06-24
txn: mutate map delete+insert

hunchback:
@roytman - this is a test for issue #259 - but it passes ... can you check?


roytman:
@hunchback , I don't think that the test correctly repeats issue #259. Please correct me, but the test uses 2 separate transactions, where 259 talks about mutation (delete) and insert operations in the same transaction.

PR #270, opened 2021-06-26
Race condition between notifciations and transaction return

roytman:
OVSDB requires that after transaction execution, the client gets the monitor notification and only after that the transaction response.
We had a race confition between notification created by transactions and etcd watch notifications. This PR resolves it.

### DIFF
--- a/pkg/ovsdb/handler.go
+++ b/pkg/ovsdb/handler.go
@@ -10,13 +10,14 @@ import (
 
 	"github.com/creachadair/jrpc2"
 	"github.com/go-logr/logr"
-	"github.com/ibm/ovsdb-etcd/pkg/common"
-	"github.com/ibm/ovsdb-etcd/pkg/libovsdb"
-	"github.com/ibm/ovsdb-etcd/pkg/ovsjson"
 	"github.com/lithammer/shortuuid/v3"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/concurrency"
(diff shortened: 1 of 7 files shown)


hunchback:
@roytman - is it still a WIP?

pkg/ovsdb/transact.go
```diff
@@ -572,8 +512,6 @@ func (etcd *Etcd) Clear() {
 	etcd.Else = []clientv3.Op{}
 	etcd.Res = nil
 	etcd.EventsNilCount = 0
```
hunchback, line 514: you can delete this aswell

roytman: done

roytman:
I'll work on the monitoring tests but you can merge this PR

pkg/ovsdb/handler.go
```diff
@@ -64,24 +65,27 @@ func (ch *Handler) Transact(ctx context.Context, params []interface{}) (interfac
...
+	if thereIsMonitor {
 		// we have to guarantee that a new monitor call if it runs concurrently with the transaction, returns first
 		var wg sync.WaitGroup
 		wg.Add(1)
-		monitor.notify(txn.etcd.Events, rev, &wg)
+		monitor.tQueue.endTransaction(rev, &wg)
 		wg.Wait()
 	}
 
+	ch.db.DbUnlock(ovsReq.DBName)
```
hunchback, line 83: rather use defer

PR #288, opened 2021-07-15
Trx cleanup

roytman:
remove unused methods
move etcd event print-related methods to database.go
allow combinations of "Select" and data unmodified operations
extend `TestTransactComment` to check the comment data
add utittest with combination of `Select` and `Comment`

### DIFF
--- a/pkg/ovsdb/database.go
+++ b/pkg/ovsdb/database.go
@@ -3,12 +3,15 @@ package ovsdb
 import (
 	"context"
 	"encoding/json"
-	"github.com/google/uuid"
-	"github.com/ibm/ovsdb-etcd/pkg/types/_Server"
+	"fmt"
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
+	"github.com/ibm/ovsdb-etcd/pkg/types/_Server"
+
 	"github.com/go-logr/logr"
+	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/concurrency"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


pkg/ovsdb/database.go
```diff
@@ -209,6 +212,87 @@ func (con *DatabaseEtcd) CreateMonitor(dbName string, handler *Handler, log logr
...
+	Key   common.Key
+	Value map[string]interface{}
+}
+
+func NewKeyValue(etcdKV *mvccpb.KeyValue) (*KeyValue, error) {
+	kv := new(KeyValue)
+
+	/* key */
+	key, err := common.ParseKey(string(etcdKV.Key))
+	if err != nil {
```
msherif1234, line 279: r u sure if no error that key ptr not nil or better to make sure key != nil before reference it below ?

pkg/ovsdb/database.go
```diff
@@ -209,6 +212,87 @@ func (con *DatabaseEtcd) CreateMonitor(dbName string, handler *Handler, log logr
...
+	err = json.Unmarshal(etcdKV.Value, &kv.Value)
+	if err != nil {
+		return nil, err
+	}
+
+	return kv, nil
+}
+
+func (kv *KeyValue) Dump() {
+	fmt.Printf("%s --> %v\n", kv.Key, kv.Value)
```
msherif1234, line 293: shouldn't we switch to logger pkg instead of fmt ?

pkg/ovsdb/transact.go
```diff
@@ -591,17 +494,28 @@ func (txn *Transaction) AddSchema(databaseSchema *libovsdb.DatabaseSchema) {
...
+Loop:
 	for _, ovsOp := range txn.request.Operations {
-		if ovsOp.Op == OP_SELECT {
+		switch ovsOp.Op {
+		case OP_SELECT:
 			hasSelect = true
-		} else {
-			hasOther = true
+			if hasSelect && hasWrite {
+				break Loop
```
msherif1234, line 506: hmm if u already set hasSelect to true why u are checking it and same below for hasWrite ?

pkg/ovsdb/database.go
roytman: I can add the check, but I doubt that key in an etcd event can be  nil

pkg/ovsdb/database.go
roytman: I have a better idea: there is no usage of the Dump method, so I will remove it for now

pkg/ovsdb/transact.go
roytman: just wrote the same test for the both switch cases, but you're right, fixed.

pkg/ovsdb/database.go
roytman: added the checks (just in case)

PR #293, opened 2021-07-26
fixing unix client logs

OmerBenHayun:


pkg/cmd/server/server.go
```diff
@@ -275,27 +275,43 @@ type ConnWrapper struct {
...
+func (c customUnixAddr) String() string {
+	return c.ClientName
+}
+func (c customUnixAddr) Network() string {
+	return "unix"
+}
+
 func (cw ConnWrapper) RemoteAddr() net.Addr {
-	return cw.intConn.RemoteAddr()
+	switch cw.intConn.RemoteAddr().Network() {
```
roytman, line 309: why not to check type of the intConn ?

OmerBenHayun: Thanks , Done

PR #296, opened 2021-08-03
Support "Where" flag in conditional monitoring

OmerBenHayun:
This PR is adding support "Where" flag in conditional monitoring.it comes into play mainly on `updater` method called `prepareRow`.
The types I covered on the tests are: `Int`,`float64` (a.k.a in `ovs` terms :`real`),`UUID`,`string`,`bool` and the types still need to cover are: `map` , `set` and `Enum`.

### (1) mock logger
when using `prepareRow` we eventually calling `NewCondition` (the call hierarchy: `prepareRow`->`isRowAppearOnWhere`->`NewCondition`). 
```go
func NewCondition(tableSchema *libovsdb.TableSchema, mapUUID MapUUID, condition []interface{}, log logr.Logger) (*Condition, error)
```
Notice that one of `NewCondition` parameters is `log logr.Logger`. Neither `prepareRow` parameters nor `updater` have access to a logger. Currently I creating a mock logger to avoid `nil` de referencing on `NewCondition` execution.
You can see that on `pkg/ovsdb/monitor.go` line `579-580`
```go
log := klogr.New() // TODO: propogate real loger insted of this generic one
res, err := NewCondition(u.tableSchema, nil, condition, log)
```
*Note:* a logger was added to `updater` so `prepareRow` can supply the logger to condition related calls.

### (2) uncovered column type test
As stated above, the types that our tests do not cover yet are: `map` , `set` and `Enum`.
### (3) interchangeable functions
We need to understand if ( `==`  , `lnclude` ) and  (`!=` , `exclude` ) are interchangeable or not on the original `OVS` protocol . Currently we treat them on the same way.
It is reasonable to assume that `lnclude` and `==`  are not the same on group element (like `set` or `map`).
**edit:**

as stated on the [RFC - Schema Format](https://datatracker.ietf.org/doc/html/rfc7047#section-3.2)

 `includes` is equivalent to `==` and `excludes` is equivalent to `!= ` for the following types:
- `boolean`
- `string`
-  `uuid`
- `integers` 
- `real`

`includes` is **not** equivalent to `==` and `excludes` is **not** equivalent to `!= ` for the following types:
- `map`
- `set`
###### TODO
- [ ] fix `CompareMap` function on `condition.go`
- [ ] fix `CompareSet` function on `condition.go`
- [ ] check correctness of function with schema during Condition (Monitor) creation.
### (4) meaningless functions return
We need to understand what the original `OVS` protocol is doing when it receives a function that is meaningless to a specific column and act the same way.
An example for meaningless functions - what is the return value of the functions `>`,`<`,`<=`,`>=` on a `map`? 
Currently `prepareRow` return on such cases an empty row.

*Note:* OVSDB clients return error

### DIFF
--- a/pkg/ovsdb/monitor.go
+++ b/pkg/ovsdb/monitor.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-logr/logr"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"k8s.io/klog/v2"
+	"k8s.io/klog/v2/klogr"
 
 	"github.com/ibm/ovsdb-etcd/pkg/common"
 	"github.com/ibm/ovsdb-etcd/pkg/libovsdb"
@@ -563,6 +564,45 @@ func unmarshalData(data []byte) (map[string]interface{}, error) {
 	return obj, nil
 }
 
+func (u *updater) isRowAppearOnWhere(data map[string]interface{}) (bool, error) {
+	checkCondition := func(condition []interface{}) (bool, error) {
+		log := klogr.New() // TODO: propogate real loger insted of this generic one
+		res, err := NewCondition(u.tableSchema, namedUUIDResolver{}, condition, log)
+		if err != nil {
+			return false, err
+		}
+		cond, err := res.Compare(&data)
+		if err != nil {
+			return false, err
+		}
+		return cond, nil
+	}
+	if u.mcr.Where == nil {
+		return true, nil
+	}
+	var cond bool
+	var err error
+	for i := 0; i < len(*u.mcr.Where); i++ {
+		switch condition := ((*u.mcr.Where)[i]).(type) {
+		case []interface{}:
+			cond, err = checkCondition(condition)
+		case [3]interface{}:
+			cond, err = checkCondition(condition[:])
+		case bool:
+			cond = condition
+		default:
+			return false, fmt.Errorf("wrong type %T %v should be a [3]interface{} or []interface{} or bool", condition, condition)
+		}
+		if err != nil {
+			return false, err
+		}
+		if cond == false {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
 func getAndDeleteUUID(data map[string]interface{}) (string, error) {
 	uuidInt, ok := data[libovsdb.COL_UUID]
 	if !ok {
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/lithammer/shortuuid/v3 v3.0.7
 	github.com/onsi/ginkgo v1.16.1
 	github.com/onsi/gomega v1.11.0
+	github.com/pkg/errors v0.8.1
```
roytman, line 16: why do you need this package ?

pkg/ovsdb/monitor.go
```diff
@@ -10,8 +10,10 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
```
roytman, line 13: probably this is a wrong `errors` package

pkg/ovsdb/monitor.go
```diff
@@ -562,6 +564,37 @@ func unmarshalData(data []byte) (map[string]interface{}, error) {
...
+	if u.mcr.Where == nil {
+		return true, nil
+	}
+	var cond bool
+	for i := 0; i < len(*u.mcr.Where); i++ {
+		switch reflect.TypeOf((*u.mcr.Where)[i]).Kind() {
+		case reflect.Slice:
+			condition, ok := ((*u.mcr.Where)[i]).([]interface{})
+			if !ok {
+				return false, errors.New("unvalid condition")
```
roytman, line 577: please add the condition's type. (I think it can be `array`, because we know that OVSDB condition is an array of 3 elements)

pkg/ovsdb/monitor.go
```diff
@@ -562,6 +564,37 @@ func unmarshalData(data []byte) (map[string]interface{}, error) {
...
+			}
+			log := klogr.New() // TODO: propogate real loger insted of this generic one
+			res, err := NewCondition(u.tableSchema, nil, condition, log)
```
roytman, line 580: instead of nil, I would add empty namedUUIDResolver, it can prevent NPE

pkg/ovsdb/monitor.go
```diff
@@ -588,12 +621,19 @@ func (u *updater) prepareRow(value []byte) (map[string]interface{}, string, erro
 	if err != nil {
 		return nil, "", err
 	}
+	res, err := u.isRowAppearOnWhere(data)
```
roytman, line 624: instead of changing the `getAdnDeleteUUID`, I would put the `isRowAppearonWhere` call in `prepareRow, before calling `getAdnDeleteUUID`

roytman (changes requested):

go.mod
OmerBenHayun: I removed the package,thanks.

pkg/ovsdb/monitor.go
OmerBenHayun: I removed the package,thanks.

pkg/ovsdb/monitor.go
OmerBenHayun: Done

pkg/ovsdb/monitor.go
OmerBenHayun: Done

pkg/ovsdb/monitor.go
OmerBenHayun: I'm calling `isRowAppearonWhere` on `prepareRow` already.
notice that `getAdnDeleteUUID` remains untouched.

PR #300, opened 2021-08-10
Support "Where" flag in conditional monitoring for set and map types

OmerBenHayun:
This PR continues #296

### DIFF
--- a/pkg/libovsdb/map.go
+++ b/pkg/libovsdb/map.go
@@ -66,3 +66,37 @@ func NewOvsMap(goMap interface{}) (*OvsMap, error) {
 	}
 	return &OvsMap{genMap}, nil
 }
+
+func IsEqualMaps(map1 OvsMap, map2 OvsMap) bool {
+	return map1.Equals(map2)
+}
+
+func (o *OvsMap) Equals(target OvsMap) bool {
+	return reflect.DeepEqual(*o, target)
+}
+
+func (o *OvsMap) containKeyValTuple(expectedKey, expectedVal interface{}) bool {
+	actualVal, ok := o.GoMap[expectedKey]
+	if !ok {
+		return false
+	}
+	return reflect.DeepEqual(expectedVal, actualVal)
+}
+
+func (o *OvsMap) IncludeMap(target OvsMap) bool {
+	for expectedKey, expectedVal := range target.GoMap {
+		if !o.containKeyValTuple(expectedKey, expectedVal) {
+			return false
+		}
+	}
+	return true
+}
+
+func (o *OvsMap) ExcludeMap(target OvsMap) bool {
+	for expectedKey, expectedVal := range target.GoMap {
+		if o.containKeyValTuple(expectedKey, expectedVal) {
+			return false
+		}
+	}
+	return true
+}
--- a/pkg/libovsdb/set.go
+++ b/pkg/libovsdb/set.go
@@ -92,3 +92,38 @@ func (o *OvsSet) UnmarshalJSON(b []byte) (err error) {
 		return addToSet(o, inter)
 	}
 }
+
+func IsEqualSets(set1 OvsSet, set2 OvsSet) bool {
+	return set1.Equals(set2)
+}
+
+func (o *OvsSet) Equals(target OvsSet) bool {
+	return len(o.GoSet) == len(target.GoSet) && o.IncludeSet(target)
+}
+
+func (o *OvsSet) containElement(expectedVal interface{}) bool {
+	for _, actualVal := range o.GoSet {
+		if reflect.DeepEqual(expectedVal, actualVal) {
+			return true
+		}
+	}
+	return false
+}
+
+func (o *OvsSet) IncludeSet(target OvsSet) bool {
+	for _, targetElement := range target.GoSet {
+		if !o.containElement(targetElement) {
+			return false
+		}
+	}
+	return true
+}
+
+func (o *OvsSet) ExcludeSet(target OvsSet) bool {
+	for _, targetElement := range target.GoSet {
+		if o.containElement(targetElement) {
+			return false
+		}
+	}
+	return true
+}
--- a/pkg/ovsdb/condition.go
+++ b/pkg/ovsdb/condition.go
@@ -203,25 +203,22 @@ func (c *Condition) CompareReal(row *map[string]interface{}) (bool, error) {
 		return false, err
 	}
 
-	if (fn == FN_EQ || fn == FN_IN) && actual == expected {
-		return true, nil
+	switch fn {
+	case FN_EQ, FN_IN:
+		return actual == expected, nil
+	case FN_NE, FN_EX:
+		return actual != expected, nil
+	case FN_GT:
+		return actual > expected, nil
+	case FN_GE:
+		return actual >= expected, nil
+	case FN_LT:
+		return actual < expected, nil
+	case FN_LE:
+		return actual <= expected, nil
+	default:
+		return false, nil
 	}
-	if (fn == FN_NE || fn == FN_EX) && actual != expected {
-		return true, nil
-	}
-	if fn == FN_GT && actual > expected {
-		return true, nil
-	}
-	if fn == FN_GE && actual >= expected {
-		return true, nil
-	}
-	if fn == FN_LT && actual < expected {
-		return true, nil
-	}
-	if fn == FN_LE && actual <= expected {
-		return true, nil
-	}
-	return false, nil
 }
 
 func (c *Condition) CompareBoolean(row *map[string]interface{}) (bool, error) {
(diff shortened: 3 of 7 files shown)


pkg/ovsdb/transact.go
```diff
@@ -45,7 +45,7 @@ const (
 )
 
 func isEqualSet(expected, actual interface{}) bool {
-	return reflect.DeepEqual(expected, actual)
+	return len(expected.(libovsdb.OvsSet).GoSet) == len(actual.(libovsdb.OvsSet).GoSet) && isIncludesSet(expected, actual)
```
roytman, line 48: why did you repalced the golang DeepEqual implementation ?

pkg/ovsdb/transact.go
```diff
@@ -45,7 +45,7 @@ const (
...
 }
 
 func isIncludesSet(expected, actual interface{}) bool {
```
roytman, line 51: The current implementation compares "==" and "include"  as the same. It was one of our long time ago misconceptions. I think we can change this and for "include" Maps and Sets use "isIncludesSet" , "isIncludemap" instead of "isEqual.."

pkg/ovsdb/condition.go
```diff
@@ -317,12 +317,77 @@ func (c *Condition) CompareEnum(row *map[string]interface{}) (bool, error) {
...
+		return libovsdb.OvsSet{GoSet: set}, nil
+	default:
+		return libovsdb.OvsSet{}, fmt.Errorf("extractSetFromRow: unexpected return type %T from extractInterfaceFromRow (return value: %v)", set, set)
+	}
+}
+
+func extractMapFromRow(row *map[string]interface{}, column string) (libovsdb.OvsMap, error) {
+	mapFromCol, err := extractInterfaceFromRow(row, column, "map")
+	if err != nil {
+		return libovsdb.OvsMap{}, err
```
roytman, line 367: you can return a pointer to map, so in the case of error it will be `return nil, err`

pkg/ovsdb/condition.go
```diff
@@ -317,12 +317,77 @@ func (c *Condition) CompareEnum(row *map[string]interface{}) (bool, error) {
...
+		return actualInterface, nil
+	default:
+		return emptyInterface, fmt.Errorf("unsupported column data type %T (column data: %v)", colData, colData)
+	}
+}
+
+func extractSetFromRow(row *map[string]interface{}, column string) (libovsdb.OvsSet, error) {
+	setFromCol, err := extractInterfaceFromRow(row, column, "set")
+	if err != nil {
+		return libovsdb.OvsSet{}, err
```
roytman, line 349: the same as for Map, you can return a pointer

pkg/ovsdb/condition.go
```diff
@@ -317,12 +317,77 @@ func (c *Condition) CompareEnum(row *map[string]interface{}) (bool, error) {
...
+	}
+	switch set := setFromCol.(type) {
+	case string:
```
roytman, line 352: you check only string, but the set can contain Intergers, Reals, and even UUIDs.
Why do you need this marshalling ?

roytman (changes requested):

pkg/ovsdb/transact.go
OmerBenHayun: Because the underlying tpe of `OvsSet` is `[]interface`.
- the order is important on go slices. 
- the order is not important on `OvsSet` (because it's a set). 

Therefore, the naive `reflect.DeepEqual` will check if the slices are equal and will take into account order of the elements on the comparison.

pkg/ovsdb/transact.go
OmerBenHayun: you are right, `==` and `include` are not the same. I updated the code.
also `!=` and `exclude` are not interchangeable anymore  on `OvsMap` and `OvsSet` and I've update it too.

pkg/ovsdb/condition.go
OmerBenHayun: this function was removed.
(we don't need it because we now using the `libovsdb.Row` unmarshaller on `prepareRow` )

pkg/ovsdb/condition.go
OmerBenHayun: same as above:
this function was removed.
(we don't need it because we now using the `libovsdb.Row` unmarshaller on `prepareRow` )

pkg/ovsdb/condition.go
OmerBenHayun: now I'm checking if the set  is equal to various of types.
currently I'm using the following cases on `CompareSet`:
```go
...
    switch data := (*row)[c.Column].(type) {
    case libovsdb.OvsSet:
        actual = data
    case int, float64, bool, string, libovsdb.UUID, libovsdb.OvsMap:
        actual = libovsdb.OvsSet{GoSet: []interface{}{data}}
    default:
        err = errors.New(E_CONSTRAINT_VIOLATION)
        c.Log.Error(err, "failed to convert row value", "value", (*row)[c.Column])
        return false, err 
    }   
...
```

pkg/ovsdb/condition.go
```diff
@@ -358,10 +369,16 @@ func (c *Condition) CompareMap(row *map[string]interface{}) (bool, error) {
 		return false, err
 	}
 
-	if (fn == FN_EQ || fn == FN_IN) && isEqualMap(actual, expected) {
```
roytman, line 361: the same comment about a switch

roytman (commented):
great job!!!
I added several comments, we can discuss them tomorrow.

pkg/libovsdb/map.go
```diff
@@ -66,3 +69,66 @@ func NewOvsMap(goMap interface{}) (*OvsMap, error) {
 	}
 	return &OvsMap{genMap}, nil
 }
+
+func IsEqualMaps(map1 *OvsMap, map2 *OvsMap) bool {
+	return map1.Equals(map2)
```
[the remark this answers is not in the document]
OmerBenHayun: interesting,thanks!
Done.

pkg/ovsdb/condition.go
```diff
@@ -333,10 +338,16 @@ func (c *Condition) CompareSet(row *map[string]interface{}) (bool, error) {
 		return false, err
 	}
 
-	if (fn == FN_EQ || fn == FN_IN) && isEqualSet(actual, expected) {
```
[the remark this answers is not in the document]
OmerBenHayun: sure, it's more elegant with a switch.
Done

pkg/ovsdb/condition.go
OmerBenHayun: sure, it's more elegant with a switch.
Done

pkg/libovsdb/set.go
```diff
@@ -92,3 +92,38 @@ func (o *OvsSet) UnmarshalJSON(b []byte) (err error) {
 		return addToSet(o, inter)
 	}
 }
+
+func IsEqualSets(set1 *OvsSet, set2 *OvsSet) bool {
```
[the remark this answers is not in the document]
OmerBenHayun: Done.

pkg/libovsdb/map.go
```diff
@@ -66,3 +69,66 @@ func NewOvsMap(goMap interface{}) (*OvsMap, error) {
...
+}
+
+func (o *OvsMap) Equals(target *OvsMap) bool {
+	//TODO make sure it's working. Assume non nil pointers
+	return reflect.DeepEqual(*o, *target)
+}
+
+func (o *OvsMap) containKeyValTuple(expectedKey, expectedVal interface{}) bool {
```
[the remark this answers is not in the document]
OmerBenHayun: removed `SplitAndSortStrings` function. now we treat maps as maps and this works as expected. but we still need the outer `containKeyValTuple` we use it on `IncludeMap` and on `ExcludeMap`.

pkg/libovsdb/map.go
```diff
@@ -66,3 +69,66 @@ func NewOvsMap(goMap interface{}) (*OvsMap, error) {
...
+			(*expectedVal) = splitAndSort(expectedValStr)
+			(*actualVal) = splitAndSort(actualValStr)
+		}
+	}
+	actualVal, ok := o.GoMap[expectedKey]
+	if !ok {
+		return false
+	}
+
+	// the following is a due to discovering that some map values
```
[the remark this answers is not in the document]
OmerBenHayun: see comment above

pkg/ovsdb/condition.go
```diff
@@ -319,8 +319,13 @@ func (c *Condition) CompareEnum(row *map[string]interface{}) (bool, error) {
 
 func (c *Condition) CompareSet(row *map[string]interface{}) (bool, error) {
 	var err error
-	actual, ok := (*row)[c.Column].(libovsdb.OvsSet)
-	if !ok {
+	var actual libovsdb.OvsSet
+	switch data := (*row)[c.Column].(type) {
+	case libovsdb.OvsSet:
+		actual = data
```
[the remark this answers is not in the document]
OmerBenHayun: we are using here [type switches](https://tour.golang.org/methods/16) . so on line `325` data is from type `libovsdb.OvsSet`. and therefore `actual` and `data` are from the same type hence can set `actual = data`.

roytman:
what is the file ` pkg/ovsdb/__debug_bin`?

OmerBenHayun:
> 
> 
> what is the file ` pkg/ovsdb/__debug_bin`?

accidentally  was included in the git commit. anyway, it has been removed.

